PR #5, opened 2021-07-30
Initial codes

jialez0:


attestation_agent/src/grpc_service.rs
```diff
@@ -0,0 +1,99 @@
+// Copyright The attestation-agent Authors.
+// SPDX-License-Identifier: Apache-2.0
+extern crate string_error;
```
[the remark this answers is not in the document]
larrydewey: This is true for all instances.

larrydewey (commented):
Just some initial review points. I will spend some more time tomorrow digging into the implementation.

attestation_agent/src/kbc_modules/sample_kbc.rs
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    fn decrypt_payload(&mut self, messages: &str) -> Result<Vec<u8>, Box<dyn Error>> {
+        // Parse the annotation, and obtain the access information of KBS and the field content to be decrypted.
+        let annotation_packet: AnnotationPacket = serde_json::from_str(messages)?;
+        self.kbs_info.push(annotation_packet.key_url);
+        self.encrypted_payload = annotation_packet.wrapped_key;
+
+        let cipher_text: &Vec<u8> = &self.encrypted_payload;
+        let decrypting_key = Key::from_slice(b"passphrasewhichneedstobe32bytes!");
```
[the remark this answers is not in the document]
voidC1326: At the current stage, the sample KBC module uses hard coded KEK for decryption. In the formal scenario, the KBC module needs to parse the annotation field passed by ocicrypt-rs, obtain the connection address information of the key broker service (KBS) and the ID of the KEK, and then communicate with KBS to actually decrypt the payload (Image encryption key) in the annotation field.

attestation_agent/src/kbc_modules/sample_kbc.rs
```diff
@@ -0,0 +1,63 @@
...
+        let cipher = Aes256Gcm::new(decrypting_key);
+        let nonce = Nonce::from_slice(b"unique nonce");
```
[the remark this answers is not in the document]
voidC1326: Please refer to the above statement.

src/grpc/mod.rs
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2021 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use futures::channel::oneshot;
+use futures::executor::block_on;
+use futures::prelude::*;
+use grpcio::{Environment, ServerBuilder};
```
fitzthum, line 9: Any particular reason to choose grpcio over tonic?

src/grpc/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+use grpc_keyprovider::keyprovider_grpc::*;
+use grpc_keyprovider::keyprovider_structs::*;
+
+use super::kbc_manager;
+
+#[derive(Clone)]
+pub struct KeyProvider;
+
+impl KeyProviderService for KeyProvider {
+    fn un_wrap_key(
```
fitzthum, line 29: Seems like unwrap should be one word? I guess this is autogenerated somewhere?

src/grpc/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+            Ok(name) => name,
+            Err(e) => {
+                println!("Some error happend when get kbc name from Dc: {}", e);
+                return ();
+            }
+        };
+
+        // Select the KBC module instance,
+        // and pass the annotation to the KBC module for content parsing and field decryption.
+        let mut kbc_manager = match kbc_manager::KBC_MANAGER.lock() {
```
fitzthum, line 61: Do we really need a mutex to unwrap a key? Will the state of the KBC or KBC Manager change?

src/grpc/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+        // Deserialize gRPC input to get kbs_annotation_packet.
+        let (annotation, dc) =
+            match get_annotation_and_dc_from_input(req.KeyProviderKeyWrapProtocolInput) {
+                Ok((a, d)) => (a, d),
+                Err(e) => {
+                    println!(
+                        "Some error happend when get annotation and dc from input: {}",
+                        e
+                    );
+                    return ();
```
fitzthum, line 44: Is there another way to do this error handling? Do we want to propagate the error back to ocicrypt? Is there a more succinct way than a match? Maybe we should at least have the message point out that the error is coming from the Attestation Agent.

src/kbc_manager.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+        ()
+    }
+
+    pub fn call_kbc_decrypt(
+        &mut self,
+        kbc_name: &str,
+        messages: &str,
+    ) -> Result<Vec<u8>, Box<dyn Error>> {
+        let kbc_instance = match self.kbc_instance_list.get_mut(kbc_name) {
```
fitzthum, line 60: I guess this allows us to use a different KBC for each wrapped key. Will we ever actually do this? Could we eliminate the KBC Manager entirely by only initializing one KBC? We will probably want to reuse encrypted container images with different KBCs. Do we want to have the KBC specified as part of the image? Another approach would be to have the AA detect which platform it is running on and choose the KBC appropriately.

fitzthum (commented):
Sorry for the delay in reviewing this. It would be nice to get the base squared away so we can start implementing KBCs. I just have a few comments.

src/grpc/mod.rs
voidC1326: AA does not need to use the additional functions provided by tonic, only using grpcio is enough, which can reduce AA footprint. On the other hand, the current grpc version is a transitional scheme, which will be replaced with lighter ttrpc along with ocicrypt-rs.

src/grpc/mod.rs
voidC1326: Yes, the name here is automatically generated by the protobuf compiler.

src/kbc_manager.rs
voidC1326: We set the KBC instance created after AA runs as the compile time option. You can either compile and support only one KBC, or specify several different kbcs at compile time.  AA will create and instanciate each and every KBC module it supports (The supported list is an build time option) when it runs. The first init call should be triggered by the first corresponding ocicrypt-rs request AA gets for a given KBC.. This solution can flexibly support various situations, and hand over the type control of KBC to the compiler (user) of AA.

src/grpc/mod.rs
voidC1326: At present, the communication protocol between AA and ocicrypt does not have a field for transmitting error information. We have put forward this suggestion to the development team of ocicrypt. They said that the communication protocol between AA and ocicrypt will be modified in the future to increase the field for transmitting error information. The ocicrypt of both the golang version and the rust version will be modified accordingly. AA will improve our error handling system after the upstream protocol is modified.

src/grpc/mod.rs
voidC1326: Thank you for reminding of this. It is necessary to use mutex, because the supported KBC instances will be created and registered in the management list of KBC manager during AA startup, resulting in the change of KBC manager. To avoid the risk of deadlock, we changed Mutex<KbcManager> to Arc<Mutex<KbcManager>>.

src/grpc/mod.rs
fitzthum: I guess this is because the protobuf says UnWrapKey rather than UnwrapKey. Looks like ocirypt is a bit inconsistent in the naming. Not a big deal.

src/kbc_manager.rs
fitzthum: I think there are some issues. Most importantly, I'm not sure it's a good idea to bake the KBC into the container image via a layer annotation. That means that a container image can only be used with one KBC, which most likely means you need different container images to run on different platforms. That's not ideal. Am I missing something?

Less of an issue, but a possible simplification, is to only support one KBC per AA. That would eliminate the need for the KBC manager. Why do we need to support multiple at once?

To account for both of these things, you could automatically select one KBC at startup and use that to fulfill all of the keyunwrap requests. Basically the AA would recognize that it's on a certain platform and use the correct KBC for that platform. This could be configurable, of course.

voidC1326: The information "which KBC to select" is generated by ocicrypt and embedded in the DC field of ocicrypt AA protocol; The KBS information (such as IP address, URL, etc.) is written in the annotation field of the container image and specified by the user encrypting the image.

On the other hand, AA supports different kbcs not to support different hw-tee platforms, but to support different kbc-kbs communication protocols (such as isecl, GOP, etc.). Since AA cannot predict which KBS is specified in the annotation in the container image，it is necessary to instantiate the KBC corresponding to all possible KBS at runtime of AA (The KBC‘s supported list is an build time option). FYI: @hdxia

fitzthum: But should the KBC information really be a property of the container image? I think it should be a property of the pod. The annotation for each layer could contain the key id, but the information about which KBC to use, the KBC URL, etc. would be in a Kubernetes annotation for the pod. As long as you provide a KBC to the pod that can provide a key matching the key id you can decrypt the container. Images shouldn't have a dependency on using a particular KBC.

voidC1326: Please see the figure below. 
截屏2021-09-06 上午10 52 10

The user encrypts his container image through the key broker service (KBS), records the KBS and encryption operation information (eg: URL, key ID, etc.) in the annotation field of the container image, and then uploads the image to the public network. After the kata agent performs the pull image operation, it needs to ask the attention agent for help to decrypt the image. At this time, the attention agent will parse the content in the annotation sent by the kata agent, access the key broker service according to the KBS information (URL) recorded therein, finally obtain the KEK, and decrypt the container image layer.

Therefore, KBS URL, key ID and other information need to be recorded in the annotation field of the container image, while KBC information does not need to exist in the container image. KBC information is a kubernetes annotation for the pod. @jiazhang0

voidC1326: Thanks for the comment. I think we may have some deviations in some terms. KBS: remote key broker service, KBC: AA's KBC module. KBS information needs to exist in the container image, but KBC information will not be stored in the container image

liangzhou121: @fitzthum here are ours answers about your questions.
1. For quesion: Could we eliminate the KBC Manager entirely by only initializing one KBC?
We think AA needs this KBC manager to support differenct type KBS platforms. As exapmle:
Intel SecL KBS (which supports TDX and SGX simultaneously):
It uses common HW-TEE remote attestation protocol to: 
  1>verify the HW-TEE's identity
  2>exchange a secret key between KBS and KBC
So a trusted and security channel can be built after Remote attestation finished. After that ISecL's specific KBC will send a getKey command to BKS to get corresponding key. This getKey command is ISecL's private protocol command.

Alibaba's EAA KBS (which supports TDX, SGX and SEV simultaneously):
It uses common HW-TEE remote attestation protocol to build a confidential and security TLS channel between KBS and KBC. This special TLS technology is based on Intel Lab's paper, the address is: https://arxiv.org/ftp/arxiv/papers/1801/1801.05863.pdf  (If you have any interesting about this).
After this TLS channel is built, EAA KBC can send getKey, unwrapKey .... commands to KBS to execute some expected operations.
These getKey, unwrapKey commands are all EAA KBS's private commands.

The upper two KBSs all use a common remote attestation protocol to check HW-TEE's identity and build a trust and security channel.
After that, it will use this channel to send KBS's private protocol commands.

So we think we can't use a single remote attestation protocol supported KBC instance, because this KBC can't support potential KBS's private protocol.


2. For question: Do we want to have the KBC specified as part of the image?
We think it's needed. 
The reason is there maybe new KBSs which supports SGX/TDX/SEV simultaneously will be developed in future. If AA can't get exact KBC's name info,  it can't choose exactly KBC.

3. For suggestion: Another approach would be to have the AA detect which platform it is running on and choose the KBC appropriately.
We think maybe we can't use this approach, becasue there maybe exist different KBSs which support SGX/TDX/SEV simultaneously. So only the platform type info isn't enough to choose the correctly KBC.

fitzthum: Ok, forget about automatic detection of the KBC for now. I meant something slightly different by that which I will explain later. There are really two separate questions. 

1. Are images portable? In your diagram the keyid and the KBS URL are part of the image annotation. Obviously the key id has to be part of the annotation because it is specific to the layer or at least could be. The question is whether we need to have KBS-specific information (like the URL) as part of the annotation. If we do, then the image would seemingly only be usable with one KBS. It is still a bit unclear where the KBC information enters the picture, but let's just focus on the KBS URL for now. Could the KBS URL be a parameter of the pod rather than the image? This would mean that each pod would have to use the same KBS, but that you could switch KBS (either to a different type of KBS or to a similar one at a different location) at any time without updating the image. The KBS URL would be a Kubernetes annotation for the pod. You could point a pod to any KBS and as long as that KBS was able to provide a key given a keyid, you could decrypt the image.

2. Do we need the KBC manager? This is a lower priority. The question is whether we need more than one KBC per pod. Obviously we will need to support more than one KBC overall because there will be different protocols. The KBC manager is basically a level of indirection that allows us to instantiate multiple KBCs and then choose the correct one for each request. If the KBS were a parameter of the pod, we would only have one KBC per pod and we seemingly would not need the KBC manager. Even if we have the KBS URL in the annotation, are users really going to need different KBCs in one pod? If not, it seems like we could simplify by just allowing them to configure the pod to use one KBC which would field every request. (As an aside, what I was thinking with automatic detection is that the user would specify one KBC or each platform in the pod config and then the AA would detect the platform and use the corresponding KBC). 

What is the status of the EAA KBS? Is it open source? Is it going to be?

liangzhou121: Hi @fitzthum,
OK, let's align whether these KBS URL should be set as pod's configuration firstly.
I think it's workable to set URL to pod. And these info maybe can be passed to AA via key provider protocol.
But put these KBS URL info into Pod's configuation file will make this design unscalable. As example, if the image's each layer was encrypted by different user via different KBS, and each layer's encrypted key is also different. Our KBC manager design want to handle this complicated scenario. Following is our design's arch:
image
In order to support this scenario, each pod need several KBCs simtaneously.

Yes, the EAA KBS is open source. And it's been developing now. It's code is in inclavare-containers project: https://github.com/alibaba/inclavare-containers

sameo: I agree with @fitzthum that things would be more portable when the KBS URL comes from the pod, i.e. this is something the kata agent should get from the host and then pass that to the AA through e.g. the key unwrap parameters. Longer term we could decide we want to support a model where if the image layers come with KBS URLs, then this gets precedence over the pod specification. Then the image non portability becomes an image owner decision. But first I think we should assume the KBS URL(s) should come from the pod spec.

Now, on the KBC modularity and detection, I first have a question for you @voidC1326 and @liangzhou121: Let's say AA is trying to get a key and got the KBS URL for that key from ocicrypt. Then how does it know which KBC to use for this key? In other words, how does it map a KBS URL (The only information it has) to a KBC? How would the AA map e.g. `https://my-cloud.com/tenant-foo/kbs` to an isecl or eaa KBC?

liangzhou121: Hi @sameo 
- Is it possable to get KBC selection info also from pod's configuration? Like  **iscel**::https://my-cloud.com/tenant-foo/kbs ? Because users know the KBS address they used, so suppose they also know the KBS type.
- Or if AA can't get any KBC info from pod, it maybe need to poll and try all supported KBC module until it can get decryption key's **Kid**'s corresponding key successfully by one KBC.

sameo: Hi @liangzhou121!

We could get it from a separate pod annotation, yes. I would not stuff it into the KBS URL annotation though. I wonder what are the security implications though, as the pod spec is controlled by the host. There may be cases where the host could use that information to try and establish secure links between a pod and a KBS that would support multiple protocols for example.

In your diagram you seem to imply that this information could also come from the agent config file. Were you thinking of mapping URLs/hosts/IPs to KBC names? The config file being controlled by the guest owner, and part of the TCB, I feel this is a safer option.

sameo: @fitzthum 

>  If the KBS were a parameter of the pod, we would only have one KBC per pod and we seemingly would not need the KBC manager

A use case that may not sound too exotic/complex: Building a guest image that could support different key broker protocols. I'm a cloud provider and I want to provide my customers with a measured guest image that could run against e.g. isecl or eaa, depending on where they choose to host their encrypted container images. I think in that case we'd need a KBC manager and allow for dynamically switching between protocols, depending on how a pod expects to talk to its KBS. There would only be one KBC per pod, but different pods could use different KBCs.
Does that make sense? Or are we over engineering here?

fitzthum: @liangzhou121

> I think it's workable to set URL to pod. And these info maybe can be passed to AA via key provider protocol or some else way. Do you have any suggestion about how to pass these info to AA?

One option is to set the KBS URL in a Kubernetes annotation. This would probably end up being part of the kernel commandline. The URL could also be part of the agent config file. Both of these are measured, which may or may not be a good thing. On the one hand, including the KBS URL in the launch measurement guarantees that it is correct. On the other hand, it increases the burden for changing the KBS URL (guest owner must recalculate measurement). It's not totally clear if the KBS URL needs to be measured for security purposes. If a malicious host changes the KBS URL the images won't decrypt.
> 
> Secondly, whether need to support more than one KBC in per pod.
> Yes, we think it's needed because it will enhance this design's scalability. As example, if the image's each layer was encrypted by different user via different KBS, and each layer's encrypted key is also different.

This is an interesting use-case. The KBC manager does make things more flexible and it isn't super complex, so I'm not totally against it. I'm not sure how common it would be for people to actually use images like this. Ultimately the layers are all being decrypted by the same user, so the two KBSes have some level of mutual trust. Maybe the keys could be consolidated to one KBS?

This approach would cause some issues with the one-URL-per-pod approach. If different layers are going to use different KBSes, different layers might also need different KBS URLs. That takes us back to the container images not being very portable.

liangzhou121: Hi @sameo @fitzthum 
We agree on your suggestions about KBS URLs and KBC names passed from Pod annotation or kata-agent config file. And we suggest to open another issue to discuss the details about their provenances.

Here is the summarized process about how to pass KBS Urls and KBC names:
- The caller of AA is responsible for passing a specific KBS url and KBC name to AA via key provider protocol.
- After receiving them, KBC manager will dynamically select and load the specified KBC module instance (So it will align with per-KBC-per-pod/per-URL-per-pod conception).

If there is no objection, we will document well about these details in IMPLEMENTATION.md in next PR. Are there any pending issues to be addressed by this PR?

sameo: > What I mean by not having a KBC manager is really just that we wouldn't parse every unwrap request to dynamically switch between KBCs. Instead the AA would instantiate one KBC on startup and then use that to directly service every request. This would still give us lots of options for how the AA selects which KBC to use.
> 

I see. So what you're proposing would avoid having to match every request to a KBC, by picking one KBC for the whole pod lifetime. How the KBC is selected is another story.
Assuming that all container images for a given pod would use the same protocol/KBC, it does make sense to me. And I think that for now, the one single protocol/KBC per pod is a safe assumption. Until someone comes with a compelling, real use case for supporting different protocols/KBC per either container or per image layers, then I believe we should start with the simpler 1:1 protocol to pod mapping assumption. 

> I briefly mentioned one scheme that the AA could use to select KBC automatically. Basically, the user could specify one KBC per platform (TDX, SEV, etc.). Then the AA could detect the platform and use the preferred KBC. This would allow an unmodified guest image to work across platforms (although maybe not a generic one built by the CSP). The guest image would need to be built with each of the KBCs specified in the config, which isn't super efficient. Since some KBSes work with multiple platforms, maybe the platform isn't the biggest concern.
>

Given that we already have one example (eaa) that plans to support both TDX and SEV, mapping a platform to a KBC would make the guest image specific to the guest owner, but still reusable across platforms. 

>  It seems like the question is really how the AA would figure out which protocol it should use. 

Or maybe in most cases it does not need to figure it out and the information comes from the kata-agent/ocicrypt, as part of the key unwrap request parameters? I think we're looking at several options:

1. Pass the selected KBC as a pod annotation. The kata-agent would get that information and then pass it down to the AA.
2. Use a guest owner defined and measured mapping between platforms (TDX, SEV, etc) and KBC. This could be part of the kata-agent configuration file and would leave it to the kata-agent to detect the hardware platform and then select a KBC.
3. Use a guest owner defined and measured mapping between KBS hosts and KBC. This is fairly similar to the previous option, in the sense that the kata-agent would parse that information and select a KBC to be passed down to the AA.

In all 3 cases, it seems to me that the AA should only rely on what it's getting through the key provider interface, and that includes the KBC. Wether the AA decides to hard wire one KBC after the first request and only support that or if we want it to dynamically pick one with each request is almost an implementation detail. Based on the one KBC per pod assumption, I would go with instanciating one KBC for the whole pod lifetime. It's simpler and more efficient imho.

> Is it going to have to reach out to the KBS URL and negotiate the protocol?

That would be ideal, although that means this URL must then be trusted.

> Speaking of the KBS URL, if the CSP provides a generic image, where will the KBS URL be specified. If we want to measure the KBS URL, it might be hard to reconcile that with having a generic image. Do we need to measure the KBS URL?

If the KBS URL comes from a pod annotation and the host tampers with it, what could happen? The AA talks to a rogue KBS, sends a quote and could even get some rogue keys back from it. But the image layers decryption or verification would then fail, iiuc. Maybe I'm missing something here, does it make sense to you?

fitzthum: @sameo 

Maybe we should just start from option 1 given that it is the simplest. 

One thing to note is that not every KBC will actually need a KBS URL. For instance a KBC for SEV(-ES) might rely only on boot-time secret injection to get secrets. Maybe the KBS URL should be part of a more general KBC option field. Some KBCs could be passed a filesystem path where the keys are baked into the guest image instead of a KBS URL.

I think the KBS URL probably does not need to be measured. What is the mechanism for conveying the URL from the Kubernetes annotation to the Kata Agent? Will it pass through the kernel commandline (which is measured by SEV)?

liangzhou121: > One thing to note is that not every KBC will actually need a KBS URL. For instance a KBC for SEV(-ES) might rely only on boot-time secret injection to get secrets. Maybe the KBS URL should be part of a more general KBC option field. Some KBCs could be passed a filesystem path where the keys are baked into the guest image instead of a KBS URL.

I think URL have already been a general name definitation and prefer don't merge into KBC option field. Because we only have 1:1 URL and KBC name mapping.
About how the URL will be parsed and used, it should be each KBC instance's specific implementation. AA's KBC manager will pass this URL parameter to KBC instance directly during KBC module's instaniation. So it should have no impact on AA's current design.

liangzhou121: Hi @sameo @fitzthum 
As we have already reached agreement of following AA's design:
- AA receives KBS Url and KBC name from key provider protocol
- AA's KBC manager (act as KBC selector + KBC loader)  will load only one KBC instance per pod and go with this instance for the whole pod lifetime.
- per KBC instance per pod && per KBS url per pod.

So I opened an separately new issue to diccuss the how to configure and store KBS Url and KBC name mapping dedicatedly.
https://github.com/containers/attestation-agent/issues/6

voidC1326: Hi @sameo @fitzthum 
After our discussion above, I summarize the latest AA architecture and operation process as follows:

截屏2021-09-10 上午11 44 16

In the above figure, how to pass KBC name and KBS URL to Kata agent is still represented as a question mark. We have opened a new issue (#6) for discussion.

sameo: > @sameo
> 
> Maybe we should just start from option 1 given that it is the simplest.

I agree.

> One thing to note is that not every KBC will actually need a KBS URL. For instance a KBC for SEV(-ES) might rely only on boot-time secret injection to get secrets. Maybe the KBS URL should be part of a more general KBC option field. Some KBCs could be passed a filesystem path where the keys are baked into the guest image instead of a KBS URL.
> 

Or we could say that we're passing a URI, where the scheme could be `http` or `file`, and always have a `(KBC, KBS URI)` pair.

> I think the KBS URL probably does not need to be measured. What is the mechanism for conveying the URL from the Kubernetes annotation to the Kata Agent? Will it pass through the kernel commandline (which is measured by SEV)?

Right now, we don't forward any pod annotation to the agent, we only carry the OCI spec ones for each container (if there are any). I think we may want to extend the [Sandbox creation request](https://github.com/kata-containers/kata-containers/blob/main/src/agent/protocols/protos/agent.proto#L276) payload to include an optional KBC+KBS URI pair.

sameo: > Hi @sameo @fitzthum
> As we have already reached agreement of following AA's design:
> 
>     * AA receives KBS Url and KBC name from key provider protocol
> 
>     * AA's KBC manager (act as KBC selector + KBC loader)  will load only one KBC instance per pod and go with this instance for the whole pod lifetime.
> 
>     * per KBC instance per pod && per KBS url per pod.
> 

That makes sense to me. @fitzthum are you ok with that approach?

fitzthum: > > Hi @sameo @fitzthum
> > As we have already reached agreement of following AA's design:
> > ```
> > * AA receives KBS Url and KBC name from key provider protocol
> > 
> > * AA's KBC manager (act as KBC selector + KBC loader)  will load only one KBC instance per pod and go with this instance for the whole pod lifetime.
> > 
> > * per KBC instance per pod && per KBS url per pod.
> > ```
> 
> That makes sense to me. @fitzthum are you ok with that approach?

Yes this LGTM. The only minor thing is that for point 1, I'm not sure we need to extend the key provider protocol to pass the KBC name and KBS URI. The kata agent can probably just give the values to the AA when the Kata Agent starts the AA.

voidC1326: I think KBC name and KBS URL should be transmitted through keyprovider protocol for the following two reasons:

1. In fact, we do not need to extend the current keyprovider protocol. The (KBC name, KBS URL) pair can be passed through the customize parameter fields reserved in the current keyprovider protocol.

2. Passing KBC name and KBS URL through keyprovider protocol instead of AA startup parameters can simplify AA architecture and startup process, unify AA parameter transmission interface, and facilitate future maintenance.

@fitzthum Do you agree to this ?

fitzthum: @voidC1326
> 
> @fitzthum Do you agree to this ?

Not really a big deal either way. It seems like it would be a bit more direct to go from the Kata Agent to the AA (although maybe ocicrypt-rs will actually be the one starting the AA; not sure). If we pass the KBC name and KBS URI through the keyprovider protocol do we have to wait for the first unwrap request before we can setup the KBC? I'm sure it will work either way.

liangzhou121: > Not really a big deal either way. It seems like it would be a bit more direct to go from the Kata Agent to the AA (although maybe ocicrypt-rs will actually be the one starting the AA; not sure). If we pass the KBC name and KBS URI through the keyprovider protocol do we have to wait for the first unwrap request before we can setup the KBC? I'm sure it will work either way.
Yes, the KBC will be setup until the first unwrap request received. And I think this mechanism's performance impact to the first unwrap request can be ignored beacuse the KBC instance setup is simple. 
So we prefer to use key provider protocol to get KBS Url and KBC name becasue as voidC1326 mentioned it will make AA's arch simple.

voidC1326: Hi @fitzthum @sameo 
I have already updated the PR and written the results of our discussion and the current implementation description of AA into the IMPLEMENTATION.md doc. Please take a moment to review them, thanks!

fitzthum: I'll take a closer look tomorrow. One thing now is that we might want to switch KBS_URL to KBS_URI. @sameo and I discussed this a bit above, but basically some KBCs might read from a file rather than connecting over the network.

docs/IMPLEMENTATION.md
```diff
@@ -0,0 +1,97 @@
...
+
+In the implementation of AA, the KBC name and KBS URI are passed through the keyprovider protocol of ocicrypt, rather than by Kata agent in the startup phase of AA.
+
+KBC name::KBS URI is passed by the user-defined field reserved in the keyprovider protocol. The request of UnWrapKey API should be as follows: 
+
+```
+{
+    "op":"keyunwrap",
+    "keyunwrapparams":{
+        "dc":{
```
fitzthum, line 72: Maybe we should add DC and EC to the glossary. I don't think the acronyms are explained anywhere in the code.

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    fn wrap_key(
+        &mut self,
+        ctx: ::grpcio::RpcContext,
+        req: keyprovider::keyProviderKeyWrapProtocolInput,
+        sink: ::grpcio::UnarySink<keyprovider::keyProviderKeyWrapProtocolOutput>,
+    ) {
+        let input_string = String::from_utf8(req.KeyProviderKeyWrapProtocolInput).unwrap();
+        println!("Request input {}", input_string);
```
fitzthum, line 98: Is this going to print the unwrapped key to stdout? You will probably be running this in a trusted environment, but that still might not be ideal.

src/kbc_runtime.rs
```diff
@@ -0,0 +1,70 @@
...
+
+// KbcInterface is a standard interface that all KBC modules need to implement.
+pub trait KbcInterface {
+    fn check(&self) -> KbcCheckInfo;
+    fn decrypt_payload(&mut self, annotation: &str) -> Result<Vec<u8>>;
+}
+
+// KbcCheckInfo is used by KBC module instances to report their internal status to AA.
+pub struct KbcCheckInfo {
+    pub kbs_info: Vec<String>,
```
fitzthum, line 23: Is `Vec<String>` really the best way to store kbs_info? Seems like maybe we want some kind of dictionary to store different info fields?

fitzthum:
Ok, this is looking good. I made a few minor suggestions. My one final question is how we can best handle situations where the attestation fails. Basically the AA will fail to service any of the gRPC requests if it either can't connect to the KBS or if the measurement is not valid. This is kind of an ugly failure because ocicrypt and the Kata Agent will just try to keep going. Not sure if there is a better way.

liangzhou121:
> Ok, this is looking good. I made a few minor suggestions. My one final question is how we can best handle situations where the attestation fails. Basically the AA will fail to service any of the gRPC requests if it either can't connect to the KBS or if the measurement is not valid. This is kind of an ugly failure because ocicrypt and the Kata Agent will just try to keep going. Not sure if there is a better way.

I think we need to confirm whether key provider protocol supports transport these error info (like KBS connection failure) from AA to ocicrypt currently. If it can't support this, we need to extend the key provider protocol firstly to add error info transportation. Then we can send a new AA PR to support this in future.

docs/IMPLEMENTATION.md
voidC1326: OK, we have added a separate chapter in the IMPLEMENTATION.md document to describe the fields and their meanings in the keyprovider protocol.

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+
+#[derive(Clone)]
+pub struct KeyProvider;
+
+impl KeyProviderService for KeyProvider {
+    fn un_wrap_key(
+        &mut self,
+        ctx: ::grpcio::RpcContext,
+        req: keyprovider::keyProviderKeyWrapProtocolInput,
+        sink: ::grpcio::UnarySink<keyprovider::keyProviderKeyWrapProtocolOutput>,
```
sameo, line 33: It seems that grpcio generates traits for the grpc services with empty return values:

```Rust
pub trait KeyProviderService {
    fn wrap_key(&mut self, ctx: ::grpcio::RpcContext, req: super::keyprovider::keyProviderKeyWrapProtocolInput, sink: ::grpcio::UnarySink<super::keyprovider::keyProviderKeyWrapProtocolOutput>);
    fn un_wrap_key(&mut self, ctx: ::grpcio::RpcContext, req: super::keyprovider::keyProviderKeyWrapProtocolInput, sink: ::grpcio::UnarySink<super::keyprovider::keyProviderKeyWrapProtocolOutput>);                                                                                                                                                                                                                                                                                   Change Visibility to pub(crate)
}

```

Is there a way to make it generate traits that return a `Result`? And then propagate that back to the client?

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    ) {
+        // Deserialize and parse the gRPC input to get KBC name, KBS uri and annotation.
+        let input_payload: InputPayload = match parse_input(req.KeyProviderKeyWrapProtocolInput) {
+            Ok(i) => i,
+            Err(e) => {
+                println!("Some error happend when parse the input: {}", e);
+                return ();
+            }
+        };
```
sameo, line 42: Having traits returning a `Result` would allow for a more idiomatic, `map_err` based handling here and below.

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+     */
+    let parameters_list = dc.parameters.get("attestation_agent").ok_or(anyhow!(
+        "Invalid parameters: the request is not sent to attention agent!"
+    ))?;
+    let kbc_kbs_pair_byte = base64::decode(parameters_list[0].clone())?;
+    let kbc_kbs_pair = std::str::from_utf8(&kbc_kbs_pair_byte)?;
+    if let Some(index) = kbc_kbs_pair.find("::") {
+        let kbc_name: String = kbc_kbs_pair[..index].to_string();
+        let kbs_uri: String = kbc_kbs_pair[(index + 2)..].to_string();
+        println!("Get KBC_NAME:{}, KBS_URI:{}", kbc_name, kbs_uri);
```
sameo, line 160: I think we want to use the `log` crate and make that a `debug!()` call instead.

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+        return Ok(InputPayload {
+            kbc_name: kbc_name,
+            kbs_uri: kbs_uri,
+            annotation: jsonstring_annotation.to_string(),
+        });
+    } else {
+        return Err(anyhow!(
+            "Invalid parameters: invalid 'kbc_name::kbs_uri' pair format!"
```
sameo, line 168: Nit: Could we be more explicit:

```Rust
return Err(anyhow!(
            "Invalid parameters: invalid '%s' pair format", kbc_kbs_pair)
```

src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+        .value_of("socket addr")
+        .unwrap_or("127.0.0.0:44444")
+        .parse::<SocketAddr>()?;
+    println!("socket addr: {:?}", socket);
+
+    // When the gRPC compilation option is used,
+    // create a gRPC service of the key provider.
+    #[cfg(feature = "grpc")]
+    grpc::start_grpc_service(socket.ip(), socket.port())
+        .map_err(|e| anyhow!("Failed starting the graph service {}", e))?;
```
sameo, line 42: s/graph/grpc/ ?

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+    let (tx, rx) = oneshot::channel();
+    thread::spawn(move || {
+        println!("Press ENTER to exit...");
+        match io::stdin().read(&mut [0]) {
+            Ok(_) => (),
+            Err(_) => panic!("Exit command unexpect!"),
+        };
+        tx.send(())
+    });
+    let _ = block_on(rx);
```
sameo, line 193: Obviously no one is going to press a key when the attestation agent either runs as a guest service or is spawned from the kata-agent.
I think here you want to handle signals (SIGINT, SIGTERM) as this is typically what this program will receive when we're shutting it down.

So using something like the`signal-hook` crate would allow for registering a `block_on(server.shutdown())` on  either SIGINT or SIGTERM.

src/grpc/mod.rs
```diff
@@ -0,0 +1,197 @@
...
+            .map(move |_| println!("Responded successfully!"));
+        ctx.spawn(output)
+    }
+
+    fn wrap_key(
+        &mut self,
+        ctx: ::grpcio::RpcContext,
+        req: keyprovider::keyProviderKeyWrapProtocolInput,
+        sink: ::grpcio::UnarySink<keyprovider::keyProviderKeyWrapProtocolOutput>,
+    ) {
```
sameo, line 96: Again, with a `Result` we should be able to simply return `Err(UNIMPLEMENTED)`.

Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+clap = "2"
+anyhow = "1.0"
+lazy_static = "1.4.0"
+string-error = "0.1.0"
+
+[build-dependencies]
+protoc-grpcio = { version = "3.0.0", optional = true }
+
+[features]
+default = ["grpc"]
```
sameo: Should gRPC even be a feature? When will we have an attestation agent that does not run a gRPC service?

Moreover, this initial code would not do much if the grpc feature would be disabled. So I propose that we remove that feature until:
1. There is an actual use case for a grpc-less attestation agent.
2. Someone implements a CLI that would actually allow for passing some input payload through a CLI.

sameo (commented):
It's looking good, thanks @voidC1326 ! I still have a few comments and opens though.
Many thanks for your patience and work.

src/grpc/mod.rs
voidC1326: At present, grpcio crate does not support this, so after weighing and considering, I decided to use tonic crate to implement grpc service. Tonic has better robustness and performance than grpcio. In order to reduce AA footprint, ttrpc will be used instead of tonic grpc in the future.

sameo (approved):

sameo (approved):
This looks really good now @voidC1326 , thanks a lot!

sameo:
@fitzthum How does that PR look to you now?

sameo:
@fitzthum 

> Ok, this is looking good. I made a few minor suggestions. My one final question is how we can best handle situations where the attestation fails. Basically the AA will fail to service any of the gRPC requests if it either can't connect to the KBS or if the measurement is not valid. This is kind of an ugly failure because ocicrypt and the Kata Agent will just try to keep going. Not sure if there is a better way.

Since this PR now relies on tonic, any grpc request failure should be propagated back to the client at least.

fitzthum (approved):
Looks great. Thanks a lot!

PR #9, opened 2021-09-17
Add KBC Module Development Guide

voidC1326:


fitzthum (approved):
Looks good. I will probably try adding new KBC next week.

voidC1326:
> Looks good. I will probably try adding new KBC next week.

@fitzthum That sounds great! Is the KBC you plan to add aimed at the sev platform?

fitzthum:
>
> @fitzthum That sounds great! Is the KBC you plan to add aimed at the sev platform?

Yes. Specifically, we're looking into an approach where there is no persistent connection between the AA and the KBS. Rather, all the secrets needed to decrypt the image are injected at boot. I think I mentioned on the original issue that we are working on some kernel patches that will expose an injected secret via the filesystem. 

It's possible that someone will want to use a similar injection approach to provision a key that allows the AA to establish a secure connection with the KBS, but we plan to start with a simpler approach.

PR #11, opened 2021-09-22
EAA KBC module

jialez0:


src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+            let decrypted_payload = base64::decode(decrypted_payload_string)?;
+            return Ok(decrypted_payload);
+        } else {
+            return Err(anyhow!(
+                "DecryptionResponse status is OK but the data is null!"
+            ));
+        }
+    }
+
+    fn kbs_get_kek(&mut self) -> Result<()> {
```
fitzthum, line 166: It doesn't look like this function is called anywhere? I guess you are using the remote attestation approach in `kbs_decrypt_payload` where you send the wrapped key to the KBS. What is the point of this function then and what is the point of keeping a cache of the KEK if you never call this function?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+                encrypted_data: base64::encode(&self.encrypted_payload),
+                algorithm: "AES".to_string(),
+                key_length: 256,
+                iv: base64::encode(&self.iv),
+            }],
+        };
+        let trans_json = serde_json::to_string(&request)?;
+        let trans_data: &[u8] = trans_json.as_bytes();
+
+        let reciv_string: String = self.kbs_trans_and_reciv(trans_data, "Dcryption")?;
```
fitzthum, line 142: Is `Dcryption` supposed to be `Decryption`?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        if let Some(hashmap_content) = kek_hashmap {
+            for (kid, kek_string) in &hashmap_content {
+                let kek = base64::decode(kek_string)?;
+                self.kek_cache.insert(kid.to_string(), kek);
+            }
+        } else {
+            return Err(anyhow!("GetKekResponse status is OK but the key is null!"));
+        }
+        Ok(())
```
fitzthum, line 192: If it were me, I would have this fucntion return the KEK rather than just update the cache. The first thing the function would do is check the kek_cache. If the requested kid is in the cache it would immediately return it. Currently it seems like the calling function has to take care of that.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,223 @@
...
+        } else {
+            return Err(anyhow!(
+                "DecryptionResponse status is OK but the data is null!"
+            ));
+        }
+    }
+
+    fn kbs_get_kek(&mut self) -> Result<()> {
+        let request = GetKekRequest {
+            command: String::from("Get KEK"),
```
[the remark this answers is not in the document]
jialez0: This string is only an instruction sent to KBS, indicating that the purpose of this function is to obtain KEK. We think there is no need to use variables here.

src/kbc_modules/eaa_kbc/mod.rs
jialez0: EAA KBS (Verdictd) can support both decrypting payload and getting KEK. We are not sure whether EAA KBC needs to handle decryption that can only use getting KEK in the future. Therefore, it is beneficial to retain this function. In order not to have an impact, it now exists as code comments.

src/kbc_modules/eaa_kbc/README.md
```diff
@@ -0,0 +1,55 @@
+# EAA KBC module
+
+## EAA Introduction
+
+Enclave Attestation Architecture (EAA) is a general attestation architecture in the cloud native scenario. EAA uses the standard attestation process to establish a secure and trusted TLS channel between EAA KBC and EAA KBS, and uses the TLS channel to send the decryption key of the encryption container image.
+
+EAA supports many different HW-TEE standard attestation processes, e.g: TDX, SEV .etc.
```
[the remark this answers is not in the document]
haosanzi: EAA KBC doesn't plan to handle pre-attestation or offline attestation in SEV(-ES). It should be covered by another KBC instance.

src/kbc_modules/eaa_kbc/rats_tls/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+        crypto: &Option<String>,
+        attester: &Option<String>,
+        verifier: &Option<String>,
+        mutual: bool,
+    ) -> Result<RatsTls, RatsTlsErrT> {
+        let mut conf: rats_tls_conf_t = Default::default();
+        conf.api_version = RATS_TLS_API_VERSION_DEFAULT;
+        conf.log_level = RATS_TLS_LOG_LEVEL_DEBUG;
+        if let Some(tls_type) = tls_type {
+            conf.tls_type[..tls_type.len()].copy_from_slice(tls_type.as_bytes());
```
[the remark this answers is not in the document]
liangzhou121: The tls_type is hardcoded as string "tls", it isn't dynamic changed during execution.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,244 @@
...
+            }
+        } else {
+            return Err(anyhow!("GetKekResponse status is OK but the key is null!"));
+        }
+
+        Ok(result_kek_map)
+    }
+    */
+
+    fn kbs_trans_and_recv(&mut self, trans_data: &[u8], error_info: &str) -> Result<String> {
```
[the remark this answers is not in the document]
liangzhou121: Yes, the KBS protocol is based on TLS socket. We didn't add length indicators to this because the currently our single package is small, so we think it won't be splited during tcp transport.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    fn establish_new_kbs_connection(&mut self) -> Result<()> {
+        debug!("create RATS TLS handle...");
+        self.tls_handle = Some(
+            rats_tls::RatsTls::new(
+                false,
+                0,
+                &Some("openssl".to_string()),
+                &Some("openssl".to_string()),
+                &Some("tdx".to_string()),
```
fitzthum, line 91: So we're hard-coding this for TDX? It's a little bit difficult to tell what each of these parameters is and it seems mybe they should be configurable somewhere?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            tls_handle: None,
+        }
+    }
+
+    fn establish_new_kbs_connection(&mut self) -> Result<()> {
+        debug!("create RATS TLS handle...");
+        self.tls_handle = Some(
+            rats_tls::RatsTls::new(
+                false,
+                0,
```
fitzthum, line 88: This is the enclave_id? Why are we always setting it to zero?

src/kbc_modules/eaa_kbc/rats_tls/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+impl DerefMut for RatsTls {
+    fn deref_mut(&mut self) -> &mut RatsTlsRef {
+        unsafe { RatsTlsRef::from_ptr_mut(self.as_ptr()) }
+    }
+}
+
+impl RatsTls {
+    pub fn new(
+        server: bool,
```
fitzthum, line 72: What does this server bool do? It seems like it should be called `is_server` or something a bit more descriptive.

fitzthum (commented):
Just a few things from me.

I don't have much experience with making C bindings, so I'm not sure if that part is optimal. It looks reasonable, though.

src/kbc_modules/eaa_kbc/mod.rs
jialez0: These parameters are the configuration parameters defined by rats-tls. For the specific meanings of these parameters, please refer to the [rats-tls project](https://github.com/alibaba/inclavare-containers/tree/master/rats-tls).

At present, we really don't have a very suitable means to configure these parameters. In [issue#7](https://github.com/containers/attestation-agent/issues/6#issuecomment-924740146), @liangzhou121 proposes to extend the keyprovider protocol and pass these KBC specific parameters to AA by ocicrypt. These KBC specific parameters can be placed in pod annotation (or somewhere else) together with KBC name and KBS URI, after this is determined, the parameters here will become configurable, but at present, we can only do the demo by hard coding these parameters.

src/kbc_modules/eaa_kbc/mod.rs
jialez0: Rats-tls first supports the remote authentication of SGX. This enclave ID is mainly used in SGX scenarios. This parameter is not used in TDX/SEV confidential computing scenarios, so it is set to zero here.

src/kbc_modules/eaa_kbc/rats_tls/mod.rs
jialez0: Since rats-tls supports two-way authentication, this bool parameter is used to indicate whether the created rats-tls instance supports server-side authentication. The name `server` should be unambiguous. It's not a big deal.

src/kbc_modules/eaa_kbc/mod.rs
fitzthum: ok. in the README this is described as a general-purpose KBC, but that doesn't fit with hard-coding TDX stuff. I guess maybe it is fine temporarily.

Do you think the KBC could automatically select the attester? For example could it check if is a TDX-enabled VM and then always use the TDX attester if it is?

In the other repo, we were discussing moving the KBC configuration to a pod annotation, which I think is a good idea in the long term. That would not work with some of these fields, however. We can't let the pod annotation select the attester or the verifier without compromising security (a rogue CSP could set verifier and attester to null). We need to be careful how we set the parameters. Doing it in the config would be secure, but it might be easier of the client if the KBC can set them automatically. 

A related question, why is the KBC selecting the verifier? Shouldn't it be up to the KBS to use whatever verifier it wants?

src/kbc_modules/eaa_kbc/mod.rs
fitzthum: ok cool. it might be handy to add a comment mentioning that server, enclave_id, and mutual are all unused.

src/kbc_modules/eaa_kbc/mod.rs
jialez0: It's a great idea to automatically detect the platform and select the attester instance. I have confirmed with the developers of rats-tls that rats-tls already has a development plan for this function, this is the [issue(#1427)](https://github.com/alibaba/inclavare-containers/issues/1427).

The verifier parameter here does not select the verifier instance of KBS. The rats-tls instance is symmetrical to the client side and the server side, the verifier parameter will only work when the `server` parameter is set to true. These parameters may cause some confusion, so as you suggested, I will add some comments on these parameters to explain.

fitzthum (approved):
Ok. I think this is good. I think we will want to add support for additional configuration and other platforms, but this seems like a good starting point.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+                result_kek_map.insert(kid.to_string(), kek.to_string());
+            }
+        } else {
+            return Err(anyhow!("GetKekResponse status is OK but the key is null!"));
+        }
+
+        Ok(result_kek_map)
+    }
+    */
+
```
sameo, line 208: Could we remove this commented out code for now?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+    pub algorithm: String,
+    pub key_length: u16,
+}
+
+pub struct EAAKbc {
+    pub kbs_uri: String,
+    pub protocol_version: String,
+    pub algorithm: String,
+    pub key_length: u16,
+    // pub kek_cache: HashMap<String, String>,
```
sameo, line 32: Can we remove that commented out code as well?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+        Ok(result_kek_map)
+    }
+    */
+
+    fn kbs_trans_and_recv(&mut self, trans_data: &[u8], error_info: &str) -> Result<String> {
+        debug!("Transmit: {}", String::from_utf8(trans_data.to_vec())?);
+        let _len_trans = self
+            .tls_handle
+            .as_ref()
+            .unwrap()
```
sameo, line 214: I'd like to avoid `unwrap`'ing as much as possible:

```rust
let _len_trans = elf
            .tls_handle
            .as_ref()
            .ok_or_else(anyhow!("Missing TLS handle"))?
            .transmit(trans_data)
            .map_err(|e| {
                error!("Transmit {} failed", error_info);
                anyhow!(format!("Something wrong when transmit, error code: {}", e))
            })?;
```

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            return Err(anyhow!("GetKekResponse status is OK but the key is null!"));
+        }
+
+        Ok(result_kek_map)
+    }
+    */
+
+    fn kbs_trans_and_recv(&mut self, trans_data: &[u8], error_info: &str) -> Result<String> {
+        debug!("Transmit: {}", String::from_utf8(trans_data.to_vec())?);
+        let _len_trans = self
```
sameo, line 211: Do we want to check that `_len_trans` is equal to `trans_data.len()`?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            })?;
+
+        // At present, the returned packets of the verdictd server is small,
+        // and the size of these packets does not exceed 4096 bytes (we have tested), 
+        // them will not be divided into multiple packets (if this happen, an error will be reported).
+        let mut buffer = [0u8; 4096];
+        let len_recv = self
+            .tls_handle
+            .as_ref()
+            .unwrap()
```
sameo, line 228: Let's avoid unwrapping here as well.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -0,0 +1,240 @@
...
+            .receive(&mut buffer)
+            .map_err(|e| {
+                error!("Transmit {} failed", error_info);
+                anyhow!(format!("Something wrong when revieve, error code: {}", e))
+            })?;
+
+        let recv_string: String = String::from_utf8(buffer[..len_recv].to_vec())?;
+        debug!("Recieve: {}", recv_string);
```
sameo, line 236: nitty nit: s/Recieve/Received/

sameo, line 236: nitty nit: s/Recieve/Received/

src/kbc_modules/eaa_kbc/rats_tls/mod.rs
sameo: Can we please remove it if we are not planning to use it?

sameo (approved):

PR #14, opened 2021-10-11
Modify sample KBC to support CCv0 PoC

jialez0:


Cargo.toml
```diff
@@ -20,6 +20,7 @@ log = "0.4.14"
 env_logger = "0.9.0"
 lazy_static = "1.4.0"
 string-error = "0.1.0"
+rust-crypto = "^0.2"
```
Jakob-Naucke, line 23: Hi @voidC1326, I work in s390x Kata development
Two things from [rust-crypto's README](https://github.com/DaGenix/rust-crypto/) worry me a bit:
> The x86-64, x86, and ARM architectures are supported

Despite this code compiling on s390x, support for s390x (and ppc64le) would be important.
> Rust-Crypto has not been thoroughly audited for correctness, so any **use where security is important is not recommended** at this time.

(emphasis mine)

The (already used) [aes-gcm](https://github.com/RustCrypto/AEADs/tree/master/aes-gcm) has neither of these issues.
Maybe this isn't a showstopper since it only appears to be used in the sample KBC at this time. But I don't think the sample should use something any practical implementation shouldn't.

src/kbc_modules/sample_kbc/mod.rs
```diff
@@ -11,18 +11,24 @@ use anyhow::*;
...
+    // This is an example annotation packet format.
+    pub kid: String,
+    pub wrapped_data: Vec<u8>,
+    pub iv: Vec<u8>,
     pub wrap_type: String,
 }
 
 pub struct SampleKbc {
     encrypted_payload: Vec<u8>,
+    iv: Vec<u8>,
```
Jakob-Naucke, line 31: Why is the IV stored in the KBC struct?

src/kbc_modules/sample_kbc/mod.rs
```diff
@@ -11,18 +11,24 @@ use anyhow::*;
...
 #[derive(Serialize, Deserialize, Debug)]
 pub struct AnnotationPacket {
-    // The access information of KBS is passed to KBC module through annotation.
-    // key_url is used as an example here.
-    pub key_url: String,
-    pub wrapped_key: Vec<u8>,
+    // This is an example annotation packet format.
+    pub kid: String,
+    pub wrapped_data: Vec<u8>,
+    pub iv: Vec<u8>,
```
Jakob-Naucke, line 25: The commit message is "Modify sample KBC to support CCv0 PoC". It wouldn't be entirely clear to me which changes are important to this. Is it these fields?

src/kbc_modules/sample_kbc/mod.rs
```diff
@@ -36,18 +41,14 @@ impl KbcInterface for SampleKbc {
...
     fn decrypt_payload(&mut self, annotation: &str) -> Result<Vec<u8>> {
-        // Parse the annotation, and obtain the access information of KBS and the field content to be decrypted.
         let annotation_packet: AnnotationPacket = serde_json::from_str(annotation)?;
-        self.encrypted_payload = annotation_packet.wrapped_key;
-
-        let cipher_text: &Vec<u8> = &self.encrypted_payload;
-        let decrypting_key = Key::from_slice(b"passphrasewhichneedstobe32bytes!");
-        let cipher = Aes256Gcm::new(decrypting_key);
-        let nonce = Nonce::from_slice(b"unique nonce");
+        self.encrypted_payload = annotation_packet.wrapped_data;
```
fitzthum, line 45: what is the point of the encrypted_payload field?

fitzthum (commented):
Looks fine in general. Have you tested the sample KBC with the sample KBS and skopeo? As in, it can actually be used to encrypt and decrypt an image?

src/kbc_modules/sample_kbc/mod.rs
jialez0: @fitzthum I have tested the following contents: encrypt a container image using skopeo and sample KBS, and decrypt the container image using skopeo and AA sample KBC. This test can run successfully.

The encrypted_payload field stores the [PrivateLayerBlockCipherOptions object](https://github.com/containers/ocicrypt/blob/main/blockcipher/blockcipher.go#L34) encrypted by KEK, which contains the necessary information and symmetric key for encryption. This field comes from the annotation of the container image layer.

Jakob-Naucke:
@jialez0 Thank you for the update! I got it to work.

PR #15, opened 2021-10-14
Add a sample KBS

jialez0:
This sample KBS is provided for testing sample KBC. It uses the same hard coded key as that in sample KBC for encryption.
Therefore, you can use this sample KBS to encrypt a container image and decrypt the image through AA's sample KBC.


jialez0:
> Ok, looks pretty good. The final thing I wonder about is whether there is any point in trying to make this more modular. I know this KBS pairs up with the sample KBC, but at some point it might be nice to provide a modular KBS implementation that could support multiple KBCs. There is also a fair amount of duplication between this and the AA. If we had a modular AA and a modular KBS, maybe they could actually be the same codebase.

Jakob-Naucke:
@jialez0 Okay, I see your point about keeping the domains separate. If we see ourselves making lots of "mirrored" changes down the road, I think we should rethink that. But it's okay for the time being. Please do keep us updated about the modularised architecture :slightly_smiling_face:! In the meantime, I shall test #17 with this PR instead.

Jakob-Naucke:
@jialez0 I understand that :) I just think it would be nice to have an encryption module correspondent to e.g. #17 without duplicating all the gRPC code. Would it also be possible to have a field for additional info (like KBCs have the KBS URI), e.g. for key ID, as sketched above?

PR #17, opened 2021-10-18
Add offline KBC

Jakob-Naucke:
Add an offline KBC that reads keys provided through a file in the guest
filesystem. This is suitable for when the filesystem is never readable
by a hypothetical adversary, e.g. with IBM Secure Execution for Linux
(IBM Z & LinuxONE). The path to the file is specified on the kernel
command line.

/cc @fitzthum @dubek

### DIFF
--- a/src/kbc_modules/offline_fs_kbc/mod.rs
+++ b/src/kbc_modules/offline_fs_kbc/mod.rs
@@ -1,0 +1,206 @@
+// Copyright (c) 2021 IBM Corp.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use crate::kbc_modules::{KbcCheckInfo, KbcInterface};
+
+use anyhow::{anyhow, Result};
+use base64::decode;
+use openssl::symm::{decrypt, Cipher};
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::fs;
+
+const KEYS_PATH: &str = "/etc/aa-offline_fs_kbc-keys.json";
+
+type Keys = HashMap<String, Vec<u8>>;
+type Ciphers = HashMap<String, Cipher>;
+
+#[derive(Deserialize)]
+pub struct AnnotationPacket {
+    // Key ID to manage multiple keys
+    pub kid: String,
+    // Encrypted key to unwrap
+    pub wrapped_data: Vec<u8>,
+    // Initialisation vector
+    pub iv: Vec<u8>,
+    // Wrap type to specify encryption algorithm and mode
+    pub wrap_type: String,
+}
+
+pub struct OfflineFsKbc {
+    // KBS info for compatibility; unused
+    kbs_info: HashMap<String, String>,
+    // Stored keys, loaded from file system; load might fail
+    keys: Result<Keys>,
+    // Known ciphers, corresponding to wrap_type
+    ciphers: Ciphers,
+}
+
+fn ciphers() -> Ciphers {
+    // The sample KBC uses aes-gcm (Rust implementation). The offline file system KBC uses OpenSSL
+    // instead to get access to hardware acceleration on more platforms (e.g. s390x). As opposed
+    // to aes-gcm, OpenSSL will only allow GCM when using AEAD. Because authentication is not
+    // handled here, AEAD cannot be used, therefore, CTR is used instead.
+    [(String::from("aes_256_ctr"), Cipher::aes_256_ctr())]
+        .iter()
+        .cloned()
+        .collect()
+}
+
+impl KbcInterface for OfflineFsKbc {
+    fn check(&self) -> Result<KbcCheckInfo> {
+        Ok(KbcCheckInfo {
+            kbs_info: self.kbs_info.clone(),
+        })
+    }
+
+    fn decrypt_payload(&mut self, annotation: &str) -> Result<Vec<u8>> {
+        let annotation_packet: AnnotationPacket = serde_json::from_str(annotation)
+            .map_err(|e| anyhow!("Failed to parse annotation: {}", e))?;
+        let kid = annotation_packet.kid;
+        let keys = self.keys.as_ref().map_err(|e| anyhow!("{}", e))?;
+        let key = keys
+            .get(&kid)
+            .ok_or_else(|| anyhow!("Received unknown key ID: {}", kid))?;
+        let iv = annotation_packet.iv;
+        let wrapped_data = annotation_packet.wrapped_data;
+        let wrap_type = annotation_packet.wrap_type;
+
+        let cipher = self
+            .ciphers
+            .get(&wrap_type)
+            .ok_or_else(|| anyhow!("Received unknown wrap type: {}", wrap_type))?;
+        // Redact decryption errors to avoid oracles
+        decrypt(*cipher, key, Some(&iv), &wrapped_data).map_err(|_| anyhow!("Failed to decrypt"))
+    }
+}
+
+impl OfflineFsKbc {
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> OfflineFsKbc {
+        OfflineFsKbc {
+            kbs_info: HashMap::new(),
+            keys: load_keys(KEYS_PATH).map_err(|e| anyhow!("Failed to load keys: {}", e)),
+            ciphers: ciphers(),
+        }
+    }
+}
+
+fn load_keys(keyfile_name: &str) -> Result<Keys> {
+    let keys_json = fs::read_to_string(keyfile_name)?;
+    // Redact parsing errors to avoid side-channels
+    let encoded_keys: HashMap<String, String> =
+        serde_json::from_str(&keys_json).map_err(|_| anyhow!("Failed to parse keys JSON file"))?;
+    encoded_keys
+        .iter()
+        .map(|(k, v)| match decode(v) {
+            Ok(key) => Ok((k.clone(), key)),
+            Err(_) => Err(anyhow!("Failed to decode key")),
+        })
+        .collect()
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use base64::encode;
+    use openssl::symm::encrypt;
+    use std::env;
+    use std::path::Path;
+
+    const KID: &str = "foo";
+    const KEY: [u8; 32] = *b"passphrasewhichneedstobe32bytes!";
+
+    #[test]
+    fn test_decrypt_payload() {
+        let iv = b"ivmustbe16bytes!";
+        let data = b"bar";
+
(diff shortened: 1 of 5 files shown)


src/kbc_modules/offline_kbc/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+        decrypt(cipher, key, Some(&iv), &wrapped_data)
+            .map_err(|e| anyhow!("Failed to decrypt: {}", e))
+    }
+}
+
+impl OfflineKbc {
+    pub fn new() -> OfflineKbc {
+        let keyfile_name = get_keyfile_name(CMDLINE)
+            .map_err(|e| anyhow!("Failed to retrieve keyfile name: {}", e))
+            .unwrap();
```
fitzthum, line 66: Maybe we should return an empty Kbc here rather than panicking. I'm not exactly sure how panicking here would affect the rest of the stack. If we return an empty Kbc, we will run into an issue later on when we try to unwrap a key, which sort of makes sense, although I'm not sure that the keyprovider interface is really built for handling errors either.

jialez0: I agree with @fitzthum , Returning an empty KBC here will cause an issue later on when we try to unwrap a key, the description information of this issue will be returned to ocicrypt through the status code of AA's grpc service, causing ocicrypt to fail to decrypt the container image layer, but AA itself will still run healthily.

src/kbc_modules/offline_kbc/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+            .ok_or_else(|| anyhow!("Received unknown key ID: {}", kid))?;
+        let iv = annotation_packet.iv;
+        let wrapped_data = annotation_packet.wrapped_data;
+
+        // TODO is the error stack a potential side channel (padding oracle attack?)
+        // and should thus not be returned?
+        // Sample uses GCM, but in OpenSSL, GCM will only work with AEAD
+        let cipher = Cipher::aes_256_ctr();
+        decrypt(cipher, key, Some(&iv), &wrapped_data)
+            .map_err(|e| anyhow!("Failed to decrypt: {}", e))
```
dubek, line 58: I don't understand why the sample can use GCM but here we can't. Can you please explain again?

src/kbc_modules/offline_kbc/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+                let result = filename.ok_or_else(|| {
+                    anyhow!(
+                        "{} was specified in kernel command line, but empty",
+                        CMDLINE_KEY
+                    )
+                });
+                return Some(result);
+            }
+            None
+        })
```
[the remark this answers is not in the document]
Jakob-Naucke: Great idea. Let's conclude the above discussion about where to specify the file first, because maybe this code will be deleted anyway.

src/kbc_modules/offline_kbc/mod.rs
Jakob-Naucke: The sample uses [RustCrypto/AEADs/aes-gcm](https://github.com/RustCrypto/AEADs/tree/master/aes-gcm). It leverages AES-NI and CLMUL on x86 and pretty much nothing else AFAICT. It will do AES/GCM with and without AEAD (AFAIU, AES/GCM without AEAD is effectively AES/CTR).
The offline KBC uses the [sfackler/rust-openssl](https://github.com/sfackler/rust-openssl) bindings. Among many other things (including x86 of course), OpenSSL also leverages CPACF on s390x, which is why I want to use it for this. It will _only_ do AES/GCM when actually using AEAD, which is why I went with AES/CTR here.
I will clarify the comment a little.

Jakob-Naucke: > I don't know about rust cryptographic libraries, but I think it's better to choose _one_ for the entire project.

I don't think OpenSSL for the entire project is a good idea because it introduces the need for OpenSSL at runtime or static linkage.
Since there isalso  no Rust AES implementation that leverages s390x hardware acceleration, would you prefer Rust's `aes-gcm`? I guess if we're only decrypting keys (not entire images) the impact isn't really big.

> I also understand that in this case there's no ditional Associated Data (AAD) you want to authenticate with the message (which must be supplied during decryption). Is that correct?

Yes; AFAIU, this is the case for all KBCs, because we manage authentication through the things standardised by container registries (GPG etc.).

src/kbc_modules/offline_kbc/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+type Keys = HashMap<String, Vec<u8>>;
+
+#[derive(Deserialize)]
+pub struct AnnotationPacket {
+    // Key ID to manage multiple keys
+    pub kid: String,
+    pub wrapped_data: Vec<u8>,
+    pub iv: Vec<u8>,
+    // Unused in offline_kbc
+    pub wrap_type: String,
```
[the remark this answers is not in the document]
Jakob-Naucke: Yes, but let's conclude the discussion in https://github.com/containers/attestation-agent/pull/17#discussion_r731595161 first.

src/kbc_modules/offline_kbc/mod.rs
Jakob-Naucke: @dubek I might add that OpenSSL has FIPS certification, which might be interesting for certification down the road. I have added the respective enablement in the PR (if we keep OpenSSL).

Jakob-Naucke: removing mandatory FIPS enable for the time being because distro support cannot be guaranteed under all circumstances. However, I'm still in favour of keeping OpenSSL.

Jakob-Naucke:
I have tested this with #14 and this diff to accomodate for things the offline FS KBC does somewhat differently:

<details><summary>diff</summary>

```diff
+++ b/sample_kbs/Cargo.toml
@@ -12,7 +12,7 @@ prost = "0.8"
 futures = "0.3.5"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
-aes-gcm = "0.9.2"
+openssl = "0.10"
 base64 = "0.13.0"
 clap = "2"
 anyhow = "1.0"
diff --git a/sample_kbs/src/crypto.rs b/sample_kbs/src/crypto.rs
index 2c209d5..e5ee715 100644
--- a/sample_kbs/src/crypto.rs
+++ b/sample_kbs/src/crypto.rs
@@ -4,8 +4,7 @@
 //
 
 use anyhow::*;
-use aes_gcm::aead::{Aead, NewAead};
-use aes_gcm::{Aes256Gcm, Key, Nonce};
+use openssl::symm::{encrypt as openssl_encrypt, Cipher};
 
 pub const HARDCODED_KEY: &'static [u8] = &[
     217, 155, 119, 5, 176, 186, 122, 22, 130, 149, 179, 163, 54, 114, 112, 176, 221, 155,
@@ -13,12 +12,9 @@ pub const HARDCODED_KEY: &'static [u8] = &[
 ];
 
 pub fn encrypt(data: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8>> {
-    let encryption_key = Key::from_slice(key);
-    let cipher = Aes256Gcm::new(encryption_key);
-    let nonce = Nonce::from_slice(iv);
-    let encrypted_data = cipher
-        .encrypt(nonce, data.as_ref())
+    let cipher = Cipher::aes_256_ctr();
+    let encrypted_data = openssl_encrypt(cipher, key, Some(iv), data)
         .map_err(|e| anyhow!("Decrypt failed: {:?}", e))?;
 
     Ok(encrypted_data)
-}
\ No newline at end of file
+}
diff --git a/sample_kbs/src/grpc/mod.rs b/sample_kbs/src/grpc/mod.rs
index 7c150c6..a0d0168 100644
--- a/sample_kbs/src/grpc/mod.rs
+++ b/sample_kbs/src/grpc/mod.rs
@@ -36,7 +36,7 @@ impl KeyProviderService for KeyProvider {
             serde_json::from_str::<KeyProviderInput>(&input_string).unwrap();
         let optsdata = input.keywrapparams.optsdata.unwrap();
 
-        let mut iv: [u8; 12] = [0; 12];
+        let mut iv: [u8; 16] = [0; 16];
         rand::rngs::OsRng.fill_bytes(&mut iv);
         let encrypt_optsdata = encrypt(
             &base64::decode(optsdata).unwrap(),
@@ -49,7 +49,7 @@ impl KeyProviderService for KeyProvider {
             kid: "null".to_string(),
             wrapped_data: encrypt_optsdata,
             iv: iv.to_vec(),
-            wrap_type: "aes-gcm".to_string(),
+            wrap_type: "aes_256_ctr".to_string(),
         };
 
         let output_struct = KeyWrapOutput {
```
</details>

and this key JSON:
```json
{
  "null": "2Zt3BbC6ehaClbOjNnJwsN2bNxv1FMqLm6fwozcR2uo="
}
```

and it works (i.e. the shasums match). Hooray.

src/kbc_modules/offline_fs_kbc/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+            keys: Ok([(
+                KID.to_string(),
+                b"thispassphraseisntactually32bytes".to_vec(),
+            )]
+            .iter()
+            .cloned()
+            .collect()),
+            ciphers: ciphers(),
+        };
+        assert!(invalid_key_kbc.decrypt_payload(&annotation).is_err());
```
dubek, line 176: The comment in line 164 says "does not yield an error" but the assert at the bottom checks that .is_err() . Which is correct?

src/kbc_modules/offline_fs_kbc/mod.rs
```diff
@@ -0,0 +1,204 @@
...
+use crate::kbc_modules::{KbcCheckInfo, KbcInterface};
+
+use anyhow::{anyhow, Result};
+use base64::decode;
+use openssl::symm::{decrypt, Cipher};
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::fs;
+
+const KEYS_PATH: &str = "/etc/aa-offline_fs_kbc-keys.json";
```
dubek, line 15: The filename doesn't have "kata" anywhere. If the attestation-agent is always part of kata-agent, it might be useful to add that. But I'm also OK with the current name.

Jakob-Naucke: I think there's no need to associate with Kata.

src/kbc_modules/offline_fs_kbc/mod.rs
dubek: Sorry, I still miss something. Why does decrypt_payload() here (line 176) return Err and not Ok(bad_decrypted_data) ?

PR #22, opened 2021-10-27
offline_fs_kbs: Do not use dd in generate-keys.sh

Jakob-Naucke:
`dd bs=32 count=1 if=/dev/random` doesn't always do what I thought it
did (the block might not actually get filled and the key will be too
short). Use `head`.


fitzthum (approved):
LGTM

sameo (approved):

PR #27, opened 2021-11-16
CI/CD: Basic build and unit tests CI.

jialez0:
Fixes:#26

### DIFF
--- a/.github/workflows/build_and_unit_tests.yml
+++ b/.github/workflows/build_and_unit_tests.yml
@@ -1,0 +1,53 @@
+name: attestation-agent build and unit tests
+on: [push, pull_request, create]
+
+jobs:
+  ci:
+    if: github.event_name == 'pull_request'
+    name: Check
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        rust:
+          - stable
+          - beta
+          - nightly
+    steps:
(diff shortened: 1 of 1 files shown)


dubek (approved):
This looks good as a first step.

Note that it only builds the default_features, which currently don't include `offline_sev_kbc` and `eaa_kbc`. 

Things to consider:
1. Add `offline_sev_kbc` to the `default_features` list in Cargo.toml
2. Add a CI step `cargo build --release --all-features` (maybe add a Makefile target for it?)
3. For (2) to work we need rats_tls installed, so maybe add a CI step to install that first.

(IMO this PR can be merged and improved later.)

jialez0:
@dubek Thank you for your advice! After consideration, I think EAA KBC and sev KBC are not suitable to be placed in the default feature because they depend on a specific running environment. Therefore, I added a separate CI to test the PR with modified EAA KBC code.

fitzthum:
I agree that the `offline_sev_kbc` and `eaa_kbc` probably shouldn't be default features. We might still want to have a basic test that tries to build all of the features, though.

I'll think if there's a good way to test `offline_sev_kbc` without specific hardware.

sameo:
Let's merge this one as we really need a basic CI. We can improve on top of it.

sameo (approved):

PR #29, opened 2021-11-16
agent: Simplify code using TryFrom trait

jodh-intel:
Simplified `parse_input()` drastically by implementing `TryFrom` for
`InputPayload` and `KeyProviderInput`.

Also refactored error handling and added lots of unit tests.

Signed-off-by: James O. D. Hunt <james.o.hunt@intel.com>

### DIFF
--- a/src/grpc/mod.rs
+++ b/src/grpc/mod.rs
@@ -5,6 +5,8 @@
 
 use anyhow::*;
 use log::*;
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
 use std::net::SocketAddr;
 use std::str;
 use std::sync::Arc;
@@ -20,8 +22,30 @@ pub mod key_provider {
     tonic::include_proto!("keyprovider");
 }
 
+pub const AGENT_NAME: &str = "attestation-agent";
+
+const ERR_ANNOTATION_NOT_BASE64: &str = "annotation is not base64 encoded";
+const ERR_DC_EMPTY: &str = "missing Dc value";
+const ERR_KBC_KBS_NOT_BASE64: &str = "KBC/KBS pair not base64 encoded";
+const ERR_KBC_KBS_NOT_FOUND: &str = "KBC/KBS pair not found";
+const ERR_NO_KBC_NAME: &str = "missing KBC name";
+const ERR_NO_KBS_URI: &str = "missing KBS URI";
+const ERR_WRONG_DC_PARAM: &str = "Dc parameter not destined for agent";
+
+const KBC_KBS_PAIR_SEP: &str = "::";
+
 #[derive(Debug, Default)]
 pub struct KeyProvider {}
+
+impl TryFrom<Vec<u8>> for InputPayload {
+    type Error = anyhow::Error;
+
+    fn try_from(bytes: Vec<u8>) -> Result<Self, Self::Error> {
+        let input = KeyProviderInput::try_from(bytes)?;
+
+        InputPayload::try_from(input)
+    }
+}
 
 #[tonic::async_trait]
 impl KeyProviderService for KeyProvider {
@@ -31,29 +55,30 @@ impl KeyProviderService for KeyProvider {
     ) -> Result<Response<KeyProviderKeyWrapProtocolOutput>, Status> {
         debug!("The UnWrapKey API is called...");
 
-        // Deserialize and parse the gRPC input to get KBC name, KBS uri and annotation.
-        let input_payload: InputPayload = parse_input(
-            request.into_inner().key_provider_key_wrap_protocol_input,
-        )
-        .map_err(|e| {
-            error!("Parse request failed: {}", e);
-            Status::internal(format!(
-                "[ERROR:attestation-agent] Parse request failed: {}",
-                e
-            ))
-        })?;
+        // Deserialize and parse the gRPC input to get KBC name, KBS URI and annotation.
+        let input_payload =
+            InputPayload::try_from(request.into_inner().key_provider_key_wrap_protocol_input)
+                .map_err(|e| {
+                    error!("Parse request failed: {}", e);
+                    Status::internal(format!(
+                        "[ERROR:{}] Parse request failed: {}",
+                        AGENT_NAME, e
+                    ))
+                })?;
 
-        // Pass the KBC name, KBS uri and annotation to the KBC instance for content parsing and field decryption.
+        // Pass the KBC name, KBS uri and annotation to the KBC instance for
+        // content parsing and field decryption.
         let kbc_runtime_mutex_clone = Arc::clone(&kbc_runtime::KBC_RUNTIME);
         let mut kbc_runtime = kbc_runtime_mutex_clone.lock().map_err(|e| {
             error!("Get KBC runtime MUTEX failed: {}", e);
             Status::internal(format!(
-                "[ERROR:attestation-agent] Get KBC runtime failed: {}",
-                e
+                "[ERROR:{}] Get KBC runtime failed: {}",
+                AGENT_NAME, e
             ))
         })?;
 
         debug!("Call KBC to decrypt...");
+
         let decrypted_optsdata = kbc_runtime
             .decrypt(
                 input_payload.kbc_name,
@@ -63,10 +88,11 @@ impl KeyProviderService for KeyProvider {
             .map_err(|e| {
                 error!("Call KBC to decrypt failed: {}", e);
                 Status::internal(format!(
-                    "[ERROR:attestation-agent] KBC decryption failed: {}",
-                    e
+                    "[ERROR:{}] KBC decryption failed: {}",
+                    AGENT_NAME, e
                 ))
             })?;
+
         debug!("Decrypted successfully, get the plain PLBCO");
 
         // Construct output structure and serialize it as the return value of gRPC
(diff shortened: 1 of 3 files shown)


jodh-intel:
I think there is more opportunity for further refactoring (and additional tests), but this is a start :smile:

jialez0 (approved):
LGTM.

jodh-intel:
@jialez0 - Thanks for reviewing. I've just updated the branch further so that now `un_wrap_key()` calls `InputPayload::try_from()` rather than `parse_input()`, as that is more idiomatic and consistent with the rest of this branch.

jodh-intel:
@sameo, @fitzthum - Please could you tal?

src/grpc/protocol/keyprovider_structs.rs
```diff
@@ -3,11 +3,24 @@
...
 
-#[derive(Serialize, Deserialize, Debug)]
+pub const ERR_INVALID_OP: &str = "invalid operator";
+pub const ERR_MISSING_OP: &str = "missing operator";
+pub const ERR_UNSUPPORTED_OP: &str = "unsupported operator";
+pub const ERR_UNWRAP_PARAMS_NO_ANNOTATION: &str = "keyunwrap parameters must include annotation";
+pub const ERR_UNWRAP_PARAMS_NO_DC: &str = "keyunwrap parameters must include dc";
+pub const ERR_WRAP_PARAMS_EXPECT_EMPTY_EC: &str = "keywrap parameters should not include ec";
+pub const ERR_WRAP_PARAMS_EXPECT_EMPTY_OPTSDATA: &str =
+    "keywrap parameters should not include optsdata";
```
fitzthum, line 19: The Attestation Agent doesn't wrap keys, hence the comments in the original that say that the ec and opts should be null. I'm not sure if this is a requirement that we need to strictly enforce, however. A keywrap request without an encryption config is in fact not a valid keywrap request. Keywrap parameters should include and ec and optsdata, you just shouldn't ever be dealing with them in the AA. 

We have talked about trying to reuse some code between the AA and the sample KBS (which does wrap keys). These plans are tentative and not a huge priority, but would require reworking this to support (and require) keywrap requests with the necessary fields.

src/grpc/protocol/keyprovider_structs.rs
```diff
@@ -17,34 +30,181 @@ pub struct KeyProviderInput {
     pub keyunwrapparams: KeyUnwrapParams,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+impl KeyProviderInput {
+    pub fn valid(&self) -> Result<()> {
+        match self.op.as_str() {
+            OP_KEY_WRAP => Err(anyhow!("{}: {:?}", ERR_UNSUPPORTED_OP, self.op)),
```
fitzthum, line 36: We might want this to be a more verbose message, suggesting that the user needs to use a different keyprovider (such as the sample kbs) to encrypt their images.

fitzthum (approved):
This looks great. 

I have one minor question regarding how we handle keywrap requests.

src/grpc/mod.rs
```diff
@@ -99,60 +125,79 @@ impl KeyProviderService for KeyProvider {
...
+
+    fn try_from(kpi: KeyProviderInput) -> Result<Self, Self::Error> {
+        let base64_annotation = kpi
+            .keyunwrapparams
+            .annotation
+            .ok_or_else(|| anyhow!(ERR_ANNOTATION_EMPTY))?;
+
+        let vec_annotation = base64::decode(base64_annotation)?;
+
+        let jsonstring_annotation: &str = str::from_utf8(&vec_annotation)?;
```
dubek, line 153: shouldn't we check at this point that the `jsonstring_annotation` is valid JSON?

src/grpc/protocol/keyprovider_structs.rs
```diff
@@ -17,34 +30,181 @@ pub struct KeyProviderInput {
...
 pub struct KeyUnwrapParams {
     pub dc: Option<Dc>,
     pub annotation: Option<String>,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+impl KeyUnwrapParams {
+    pub fn valid(&self) -> Result<()> {
+        // "empty" means "not defined" or "not set"
+        let dc_empty = if let Some(dc) = self.dc.clone() {
```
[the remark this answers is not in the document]
jodh-intel: It's because `Dc` doesn't implement `Copy` - try removing the `clone()` and the compiler will bleat at you :smile:

src/grpc/mod.rs
jodh-intel: I believe that string should be just plain text at this point. I think that ariable name is erroneous, I've renamed it to `annotation` to avoid the confusion, but it would be good to get confirmation from @jialez0.

src/grpc/protocol/keyprovider_structs.rs
jodh-intel: @fitzthum - Thanks. I've updated the message.

src/grpc/protocol/keyprovider_structs.rs
jodh-intel: @fitzthum - ok. I'm not totally clear from your comment whether you'd like me to remove that additional checking?

jodh-intel:
Branch updated with the following changes:

- global: Added new `ERR_ANNOTATION_NOT_BASE64` and `ERR_KBC_KBS_NOT_BASE64` errors.
- `InputPayload`: Simplified `try_from()`.
- `Dc`: Added in check for valid URI (and new unit tests).
- tests: Improved failure messages.
- Fixed a potential crasher bug if the Dc parameters vector was empty (I've added a unit test).

I've also found a problem: as noted above the code now checks the KBS URI to make sure it looks valid. I've added a few unit tests but then I found https://github.com/confidential-containers/attestation-agent/tree/main/src/kbc_modules/eaa_kbc#usage which includes:

> Set KBC_NAME::KBS_URI pair in the following format:
> ```
> eaa_kbc::<IP>:<PORT>
> ```
>
> e.g: EAA KBS address is 127.0.0.1:1122 :
>
> ``` 
> eaa_kbc::127.0.0.1:1122
> ```

The problem is `eaa_kbc` is not a valid URI scheme since schemes can only use alphabetics, `+`, `-` and `.` (see https://datatracker.ietf.org/doc/html/rfc3986#section-3.1).

Can I just go ahead and "fix" that document so the scheme becomes `eaa-kbc`? If so, I'll also add unit tests for `eaa_kbc` and `eaa-kbc`.

Jakob-Naucke:
> The problem is `eaa_kbc` is not a valid URI scheme since schemes can only use alphabetics, `+`, `-` and `.` (see https://datatracker.ietf.org/doc/html/rfc3986#section-3.1).
> 
> Can I just go ahead and "fix" that document so the scheme becomes `eaa-kbc`? If so, I'll also add unit tests for `eaa_kbc` and `eaa-kbc`.

Hmm. That would apply to all KBCs implemented thus far. We used underscores because Rust modules generally use snake casing and didn't really consider URL validity. We could, but users should be very aware of this change because their configurations will need updating. @fitzthum WDYT?

jodh-intel:
It would be a real shame if we didn't comply with the URI spec though :cry:

fitzthum:
We're expecting`KBC_NAME::KBS_URI`. I think only the second half is supposed to be the URI. `KBC_NAME` isn't the scheme so I think its format may be fine. That said, we aren't actually providing a URI in `KBS_URI` in that example because it has no scheme. 

Originally the idea with a URI was to differentiate KBCs that point to external resources like KBS and KBCs that point to internal resources like files. We ended up hardcoding the path for our two file-based KBCs so this feature wasn't given much scrutiny. Based on my initial understanding, we might want to update the `KBC_URI` to have a scheme.

It's also slightly weird that the `KBS_URI` is pointing to localhost in that example. That would seem to suggest that the KBS is running inside the enclave, which is odd.

src/grpc/protocol/keyprovider_structs.rs
fitzthum: I think it's technically fine as it is. I guess we could skip the valid check on the keywrapparams. We don't really need some of the related functionality or tests but it doesn't hurt.

jodh-intel:
Hi @fitzthum - thanks for explaining the context here. We still have a problem though because a URI requires a scheme, so `127.0.0.1:1122` is "invalid" as there is no scheme prefix, hence some questions:

- Do we / can we assume a default scheme to allow the URI to be validated?
- `<IP>`: Do we mean IPv4 *and* IPv6?

I'd like to be able to validate the URI but without some sort of scheme we can't rely on standard facilities.

/cc @sameo.

jialez0:
@jodh-intel I'm sorry that we didn't consider the URI format specification before. As you said, if we need to verify the URI here with the standard facilities, maybe we need to add an appropriate scheme prefix to socket address such as `<IPv4/v6>:<port>`, but I don't know the URI format specification, so I'm not sure what kind of scheme prefix it should be.

dubek:
I think this is the address for the KBS grpc endpoint. grpc is implemented over http2 if I remember correctly (but please check). If this is true, then an `https://` scheme should work:   `https://IPv4:PORT/` (not sure how ipv6 works in a URI).

jodh-intel:
Branch rebased to resolve the conflict introduced by #31.

Thanks for the comments, but I'm not convinced about the URL validation. It sounds like the topic needs further discussion to me, so I've removed that part of the branch. I'm happy to add it at a later date.

sameo (approved):

jodh-intel:
Fixed clippy test code nits.

jodh-intel:
All green ;)

PR #31, opened 2021-11-22
Fix: CI/CD failed

jialez0:

PR #32, opened 2021-11-22
CI/CD: fix: eaa_kbc CI run failed

jialez0:


.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo apt-get update;
+          sudo apt-get install -y autoconf gcc g++ make wget git cmake \
           libseccomp-dev binutils-dev libprotoc-dev protobuf-compiler \
           pkg-config libssl-dev openssl libtool file curl gnupg;
 
-          wget https://github.com/protobuf-c/protobuf-c/archive/v1.3.1.tar.gz &&\
-          tar -zxvf v1.3.1.tar.gz && cd protobuf-c-1.3.1 && ./autogen.sh && \
-          ./configure && make && make install
+          sudo wget https://github.com/protobuf-c/protobuf-c/archive/v1.3.1.tar.gz;
+          sudo tar -zxvf v1.3.1.tar.gz;
```
jodh-intel, line 34: Do you have to use this version? If so:

- Please add a comment explaining why.
- Please use a variable to avoid the duplicated version string.

... but if you just need the latest, you could use something like:

```bash
local latest
latest=$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/protobuf-c/protobuf-c/releases/latest)
```

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
-          apt-get update && apt-get install -y autoconf gcc g++ make wget git cmake \
+          sudo apt-get update;
+          sudo apt-get install -y autoconf gcc g++ make wget git cmake \
           libseccomp-dev binutils-dev libprotoc-dev protobuf-compiler \
           pkg-config libssl-dev openssl libtool file curl gnupg;
 
-          wget https://github.com/protobuf-c/protobuf-c/archive/v1.3.1.tar.gz &&\
-          tar -zxvf v1.3.1.tar.gz && cd protobuf-c-1.3.1 && ./autogen.sh && \
-          ./configure && make && make install
+          sudo wget https://github.com/protobuf-c/protobuf-c/archive/v1.3.1.tar.gz;
```
jodh-intel, line 33: - `curl` might be a better alternative from a security perspective.
- This network download is running as `root`. It would be safer to download as a non-priv user and then elevate privs to `root` if needed later.

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo tar -zxvf v1.3.1.tar.gz;
+          cd protobuf-c-1.3.1 && sudo ./autogen.sh;
+          sudo ./configure;
+          sudo make;
+          sudo make install;
```
jodh-intel, line 38: Surely only the install step needs to be run as `root`?

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
 
-          wget https://download.01.org/intel-sgx/sgx-linux/2.14/as.ld.objdump.gold.r3.tar.gz && \
-          tar -zxvf as.ld.objdump.gold.r3.tar.gz && cp -rf external/toolset/ubuntu18.04/* /usr/local/bin/ && \
-          rm -rf external && rm -rf as.ld.objdump.gold.r3.tar.gz
-
-          wget https://download.01.org/intel-sgx/sgx-linux/2.14/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.14.100.2.bin;
-          chmod +x sgx_linux_x64_sdk_2.14.100.2.bin &&  echo -e 'no\n/opt/intel\n' | ./sgx_linux_x64_sdk_2.14.100.2.bin;
-          rm -rf sgx_linux_x64_sdk_2.14.100.2.bin;
+          sudo wget https://download.01.org/intel-sgx/sgx-linux/2.14/as.ld.objdump.gold.r3.tar.gz;
+          sudo tar -zxvf as.ld.objdump.gold.r3.tar.gz;
```
jodh-intel, line 41: Same questions as above about versions here.

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo cp -rf external/toolset/ubuntu18.04/* /usr/local/bin/ && \
```
jodh-intel, line 42: Is this definitely `ubuntu18.04` and not `ubuntu-18.04` to align with the first change on this PR?

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo rm -rf external && sudo rm -rf as.ld.objdump.gold.r3.tar.gz
 
+          sudo wget https://download.01.org/intel-sgx/sgx-linux/2.14/distro/ubuntu18.04-server/sgx_linux_x64_sdk_2.14.100.2.bin;
+          sudo chmod +x sgx_linux_x64_sdk_2.14.100.2.bin;
+          echo -e 'no\n/opt/intel\n' | sudo ./sgx_linux_x64_sdk_2.14.100.2.bin;
```
jodh-intel, line 47: Might be worth adding a comment to explain what "no" means. Presumably `/opt/intel` is the prefix?

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo rm -rf sgx_linux_x64_sdk_2.14.100.2.bin;
           source /opt/intel/sgxsdk/environment;
-
-          echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main' | tee /etc/apt/sources.list.d/intel-sgx.list && \
-          wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | apt-key add -;
+          echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu bionic main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list;
```
jodh-intel, line 50: Hard-coded architecture. May be fine for now but something to think about.

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo rm /var/lib/apt/lists/lock && sudo rm /var/lib/dpkg/lock;
+          sudo apt-get --allow-unauthenticated upgrade;
```
jodh-intel, line 52: Can you add a comment explaining why `--allow-unauthenticated` is required?

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          sudo wget -qO - https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo apt-key add -;
           
-          apt-get update -y && apt-get install -y libsgx-dcap-quote-verify libsgx-dcap-default-qpl libsgx-dcap-ql-dev libsgx-uae-service libsgx-dcap-quote-verify-dev;
+          sudo apt-get update -y;
+          sudo apt-get install -y libsgx-dcap-quote-verify libsgx-dcap-default-qpl libsgx-dcap-ql-dev libsgx-uae-service libsgx-dcap-quote-verify-dev;
```
jodh-intel, line 56: It would be clearer if you used backslashes to fold each package onto a separate line (and ensured the list is sorted alphabetically). Same comment for other packages on this PR.

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
-          cmake -DBUILD_SAMPLES=on -H. -Bbuild && make -C build install;
+          git clone https://github.com/alibaba/inclavare-containers;
+          cd inclavare-containers/rats-tls;
+          sudo cmake -DBUILD_SAMPLES=on -H. -Bbuild;
+          sudo make -C build install;
       
       - name: Build AA with EAA KBC
         run: |
-          make KBC=eaa_kbc && make install;
+          ls;
```
jodh-intel, line 67: Is this debug output that can be removed or still needed?

.github/workflows/eaa_kbc.yml
```diff
@@ -25,34 +25,47 @@ jobs:
...
+          curl https://sh.rustup.rs -sSf > rust_install.sh;
```
jodh-intel, line 68: Nit: https://www.rust-lang.org/tools/install shows the command as:

```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs ...
```

jodh-intel (commented):
Thanks @jialez0.

A few comments. One general one being that it looks like you can remove a lot of the `sudo` calls.

jialez0:
@jodh-intel Thanks for review! In order to more intuitively observe the modification effect, I will merge the commit in this PR into PR #31, and I will continue to process and modify the contents mentioned in your comment in PR #31.

jodh-intel:
Thanks @jialez0.

Issue #36, opened 2021-11-26
Resolve missing URI scheme issue


jodh-intel:
Writing tests for #29 uncovered the fact that the KBS URI is not technically a URI since it doesn't include a scheme.

We need to resolve which scheme is appropriate. We're using ttrpc so is it http, https, h2c, h2, other?

A minimal diff once we've decided is something like:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 1da653a..5200469 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ lazy_static = "1.4.0"
 string-error = "0.1.0"
 foreign-types = "0.5.0"
 openssl = { version = "0.10", optional = true, features = ["vendored"]}
+url = "2.2.2"
 
 [build-dependencies]
 tonic-build = "0.5"
diff --git a/src/grpc/mod.rs b/src/grpc/mod.rs
index 51820bf..dec2e94 100644
--- a/src/grpc/mod.rs
+++ b/src/grpc/mod.rs
@@ -28,6 +28,7 @@ const ERR_ANNOTATION_NOT_BASE64: &str = "annotation is not base64 encoded";
 const ERR_DC_EMPTY: &str = "missing Dc value";
 const ERR_KBC_KBS_NOT_BASE64: &str = "KBC/KBS pair not base64 encoded";
 const ERR_KBC_KBS_NOT_FOUND: &str = "KBC/KBS pair not found";
+const ERR_KBS_URI_INVALID: &str = "KBS URI is invalid";
 const ERR_NO_KBC_NAME: &str = "missing KBC name";
 const ERR_NO_KBS_URI: &str = "missing KBS URI";
 const ERR_WRONG_DC_PARAM: &str = "Dc parameter not destined for agent";
@@ -222,6 +223,9 @@ fn str_to_kbc_kbs(value: &str) -> Result<(String, String)> {
             return Err(anyhow!(ERR_NO_KBS_URI));
         }
 
+        let _ =
+            url::Url::parse(kbs_uri).map_err(|e| anyhow!("{}: {:?}", ERR_KBS_URI_INVALID, e))?;
+
         Ok((kbc_name.to_string(), kbs_uri.to_string()))
     } else {
         Err(anyhow!(ERR_KBC_KBS_NOT_FOUND))
@@ -688,6 +692,10 @@ mod tests {
                 value: "::foo",
[...]

PR #43, opened 2021-12-21
Add `GetResource` gRPC service and endpoint.

jialez0:


protos/getresource.proto
```diff
@@ -0,0 +1,17 @@
...
+package getresource;
+
+message getResourceRequest {
+    string KbcName = 1;
+    string KbsUri = 2;
+    string ResourceDescription = 3;
+}
+
+message getResourceResponse {
+    bytes resource = 1;
```
fitzthum, line 12: resource -> Resource?

src/kbc_runtime.rs
```diff
@@ -54,6 +54,23 @@ impl KbcRuntime {
...
+        resource_description: String,
+    ) -> Result<Vec<u8>> {
+        if self.kbc_instance_map.get_mut(&kbc_name).is_none() {
+            self.instantiate_kbc(kbc_name.clone(), kbs_uri)?;
+        }
+        let kbc_instance = self
+            .kbc_instance_map
+            .get_mut(&kbc_name)
+            .ok_or(anyhow!("KBC runtime: The KBC instance does not existing!"))?;
+        let resource = kbc_instance.get_resource(resource_description)?;
```
fitzthum, line 70: How can we handle KBCs that don't have a get_resource endpoint? I think we might want to add that endpoint as a stub to every existing KBC and create some mechanism for reporting that the feature isn't supported.

src/main.rs
```diff
@@ -20,23 +20,42 @@ async fn main() -> Result<()> {
...
-                .long("grpc_sock")
+            Arg::with_name("KeyProvider gRPC socket addr")
+                .long("keyprovider_sock")
                 .takes_value(true)
-                .help("The socket address which the grpc service will listen to, for example: --grpc_sock 127.0.0.1:11223",
+                .help("This socket address which the KeyProvider gRPC service will listen to, for example: --keyprovider_sock 127.0.0.1:11223",
+                ),
+        )
+        .arg(
+            Arg::with_name("GetResource gRPC socket addr")
```
fitzthum, line 30: Why create two different sockets for this? In practice won't `ocicrypt-rs` be the one using both? I guess it might be more robust to have a separate socket?

src/grpc/keyprovider/message.rs
```diff
@@ -52,17 +52,20 @@ impl KeyProviderInput {
         Ok(())
     }
 
+    #[must_use]
```
fitzthum, line 55: You might want to put this stuff in a separate PR. Should be trivial to merge.

protos/getresource.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+package getresource;
+
+message getResourceRequest {
```
fitzthum, line 5: I am not familiar with the convention here. Why is the first letter lowercase?

fitzthum (commented):
Now that the the `image-rs` proposals have seemed to settle, let's try to get this merged. Sorry for the delay. A few thoughts to start with.

src/main.rs
jialez0: My considerations of using two separate sockets are as follows:

1. In fact, the caller of get resource is not ocicrypt RS, but image rs. ocicrypt RS does not support container image signature verification, but only image encryption and decryption. Image RS actually completes image signature verification. (this follows the implementation logic of ocicrypt and image in golang language version)

2. The protobuf of the key provider grpc service has been defined by ocicrypt. In order to maintain compatibility as much as possible, we'd better not change it.

src/kbc_runtime.rs
jialez0: In the trait definition of KBC interface, we can add the default implementation of KBC's get resource interface: https://github.com/confidential-containers/attestation-agent/pull/43/commits/4f3ae318b4fde38b637ac38bb41992987900177a 

For KBC without custom implementation of get resource interface, this default implementation will be used.

src/kbc_modules/mod.rs
```diff
@@ -24,6 +24,9 @@ use std::sync::Arc;
 pub trait KbcInterface {
     fn check(&self) -> Result<KbcCheckInfo>;
     fn decrypt_payload(&mut self, annotation: &str) -> Result<Vec<u8>>;
+    fn get_resource(&self, _description: String) -> Result<Vec<u8>> {
+        Err(anyhow!("Get Resource API of this KBC is unimplement!"))
```
fitzthum, line 28: nit: `unimplement` -> `not implemented`

fitzthum (approved):
LGTM. This PR seems fairly straightforward. I have a few minor comments. 

We'll need to adjust the Kata Agent slightly to set the additional socket. Let's remember to do that sometime.

The format of the resource description will require some thinking, but this PR doesn't determine that.

src/main.rs
```diff
@@ -20,23 +20,42 @@ async fn main() -> Result<()> {
...
+        "GetResource gRPC service listening on: {:?}",
+        getresource_socket
+    );
 
-    grpc::start_service(socket).await?;
+    let keyprovider_server = grpc::keyprovider::start_service(keyprovider_socket);
+    let getresource_server = grpc::getresource::start_service(getresource_socket);
 
-    Ok(())
+    tokio::join!(keyprovider_server, getresource_server).0
```
fitzthum, line 60: This is slightly weird. `.0` gives us the status of `keyprovider_server` only?

liangzhou121:
>The format of the resource description will require some thinking, but this PR doesn't determine that.

Hi @fitzthum ,
We have created a new issue to discuss the definition of `resource description`. Please help to have a review also.
Issue: https://github.com/confidential-containers/guest-components/issues/243

protos/getresource.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+
+package getresource;
+
+message GetResourceRequest {
+    string KbcName = 1;
+    string KbsUri = 2;
```
jodh-intel, line 7: I'd consider documenting all message fields. For example, is this a real URL or a "fake" one? See https://github.com/confidential-containers/attestation-agent/issues/36.

protos/getresource.proto
```diff
@@ -0,0 +1,17 @@
...
+    string ResourceDescription = 3;
+}
+
+message GetResourceResponse {
+    bytes Resource = 1;
```
jodh-intel, line 12: What exactly _is_ `Resource`?

src/grpc/getresource/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub mod get_resource {
+    tonic::include_proto!("getresource");
+}
+
+#[derive(Debug, Default)]
+pub struct GetResource {}
+
+#[tonic::async_trait]
+impl GetResourceService for GetResource {
```
jodh-intel, line 25: This looks odd to me. Wouldn't `Resource` be a better name for the `struct` which would then become:

```rust
impl GetResourceService for Resource
```

src/grpc/getresource/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+        &self,
+        request: Request<GetResourceRequest>,
+    ) -> Result<Response<GetResourceResponse>, Status> {
+        let request = request.into_inner();
+
+        let kbc_runtime_mutex_clone = Arc::clone(&kbc_runtime::KBC_RUNTIME);
+        let mut kbc_runtime = kbc_runtime_mutex_clone.lock().map_err(|e| {
+            error!("Get KBC runtime MUTEX failed: {}", e);
+            Status::internal(format!(
+                "[ERROR:{}] Get KBC runtime failed: {}",
```
jodh-intel, line 36: Not required for this PR, but I think we need a macro to avoid the constant use of the `[ERROR:` prefix.

jodh-intel (commented):
Thanks @jialez0. A few comments...

src/grpc/keyprovider/message.rs
jialez0: Done.

fitzthum:
Ok, I guess we are waiting for `image-rs` support to get merged?

jialez0:
\tests

jialez0:
I think this PR can get merged first and don't need to wait for image-rs, so that we can start adding KBC support for get resource :)

PR #45, opened 2022-01-25
Add a Cargo.lock file

Jakob-Naucke:
to avoid accumulating incompatible versions.

The one provided here was built with `--no-default-features`. https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries suggests binary applications should lock their versions. If you disagree, we can also use more rigorous version locks in `Cargo.toml`. What I'd like to avoid is random breakage, which happened on s390x with socket2 v0.4.3 in combination with libc v0.2.113 (see https://github.com/rust-lang/libc/pull/2645#issuecomment-1020128032).


fitzthum (approved):
Seems like a good idea.

jialez0:
@Jakob-Naucke Can you please rebase this PR and we can merge it.

Jakob-Naucke:
Thanks @jialez0 -- this Cargo.lock is from just now.

PR #46, opened 2022-01-27
Fix CI run failed.

jialez0:


.github/workflows/basic.yml
```diff
@@ -50,4 +50,4 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets --all-features -- -D warnings
\ No newline at end of file
+          args: --all-targets --all-features
```
Jakob-Naucke, line 53: TBH I can't even reproduce the `must_use` errors, even with a nightly toolchain (I can reproduce the `.to_string()` error though). But with the added `must_use`s, is it really necessary to allow warnings?

jialez0: After rustup-nightly was updated to version 1.59.0,
the lint check warns when a method returning Self doesn’t have the #[must_use] attribute.
(refer to: https://rust-lang.github.io/rust-clippy/master/index.html#return_self_not_must_use)
However, at present, the rust source code generated by proto built by tonic tool chain
does not comply with this requirement, which will lead to CI/CD running failure: 
https://github.com/confidential-containers/attestation-agent/runs/4961352636?check_suite_focus=true#step:8:110

Jakob-Naucke: Hm. Is it important we run on nightly? Because I should think tonic will have fixed this by the time 1.59 becomes stable, and then we won't have to think about re-enabling `-D warnings` (or fix additional warnings that come up by then).

Jakob-Naucke (approved):
Thanks for the updates @jialez0.

fitzthum (approved):
To me it looks like we are only having the `must_use` problem on `beta` not `nightly`, but it is probably fine to skip them both. It looks like some other CI platforms have an `allow_failure` flag that might be perfect for this but I don't think it's supported with Github actions

PR #47, opened 2022-02-28
[EAA KBC] Support `get_resource` interface.

jialez0:


src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -15,6 +15,12 @@ pub mod protocol;
 pub mod rats_tls;
 use protocol::*;
 
+const DEFAULT_RECV_SIZE: usize = 4096;
```
jodh-intel, line 18: - ooi, why this value?
- Should it be configurable (the fact the name contains "default" suggests it could be overridden somehow)?
- Please add the units to the name of this `const` for clarity (`_BYTES`?)

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -154,7 +173,77 @@ impl EAAKbc {
...
+        let buffer_size: usize = resource_info.base64size.parse::<usize>()?;
+
+        let recv_string: String =
+            self.kbs_trans_and_recv(trans_data, buffer_size, "Get Resource")?;
+
+        let data_bytes = base64::decode(recv_string)?;
+
+        if data_bytes == b"Error".to_vec() {
+            return Err(anyhow!(
+                "Verdictd: The resource data is null, something failed."
```
jodh-intel, line 207: - Why is this prefixed with "verdictd"? Should it just be "verdict"?
- What is generating this `Error` message? The code looks fragile since it would break if the output format changes.
- Why is it not possible to determine more detail on what caused the "error"?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -154,7 +173,77 @@ impl EAAKbc {
...
+                if let Some(info) = response.data {
+                    Ok(info)
+                } else {
+                    return Err(anyhow!("Verdictd: Resource info payload is null!"));
+                }
+            }
+            "Fail" => return Err(anyhow!(format!("Verdictd: {}", response.error.unwrap()))),
+            _ => {
+                return Err(anyhow!(
+                    "Can't understand the GetResourceInfoResponse status!"
```
jodh-intel, line 235: - Personally, I wouldn't include exclamations in error messages: they don't add any additional information. It's going to be clear to the admin this is an error so that is sufficient.
- I'd also avoid apostrophe's where possible to since these signify a contraction and not all reads will understand these shortened forms of words. Plus they can cause problems for log parsers / formatters.

src/kbc_modules/eaa_kbc/protocol.rs
```diff
@@ -7,6 +7,11 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::vec::Vec;
 
+pub const GET_POLICY_COMMAND: &str = "Get Policy";
+pub const GET_SIGSTORE_CONFIG_COMMAND: &str = "Get Sigstore Config";
+pub const GET_GPG_KEY_RING_COMMAND: &str = "Get GPG Keyring";
```
jodh-intel, line 12: These appear to be identical to the `const`'s in src/kbc_modules/eaa_kbc/mod.rs, but with a "`Get `" prefix. Is there any way to generate these from the `const` values in `mod.rs` to make this relationship clearer (and ensure the code doesn't break if `mod.rs` is updated.

jodh-intel (commented):
Thanks @jialez0. A few comments...

Also, is there any way you could add some new unit tests for these changes?

src/kbc_modules/eaa_kbc/mod.rs
jialez0: This is a default receiving buffer size determined after testing. It is used for some regular information interaction (such as version query). It can be overwritten. When calling it, the caller can decide whether to use this default size. It will give priority to the specified receiving buffer size.

jialez0:
@jodh-intel These new functions of EAA KBC cannot be tested in unit test for the time being, because their test needs to interact with EAA KBS (Verdictd). In the future, we can add tests on these functions in CI/CD system.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -165,23 +263,37 @@ impl EAAKbc {
...
+                .as_ref()
+                .ok_or(anyhow!("Missing TLS handle"))?
+                .receive(&mut recv_buffer)
+                .map_err(|e| {
+                    error!("Transmit {} failed", error_info);
+                    anyhow!(format!("Something wrong when revieve, error code: {}", e))
+                })?;
 
-        let recv_string: String = String::from_utf8(buffer[..len_recv].to_vec())?;
+            recv_res.append(&mut recv_buffer[..len_single].to_vec());
```
fitzthum, line 281: Is `receive` going to return something different when you call it multiple times? Otherwise it seems like you are just going to be appending the same thing to the buffer each time the loop runs.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -149,12 +174,85 @@ impl EAAKbc {
...
+
+    fn kbs_trans_and_recv(
+        &mut self,
+        trans_data: &[u8],
+        recv_size: usize,
+        error_info: &str,
+    ) -> Result<String> {
+        if trans_data.to_vec().is_empty() || recv_size == 0 {
+            return Err(anyhow!(
+                "Transmit data and recieve buffer can not be empty."
```
fitzthum, line 253: Nit: Message is slightly inaccurate. Receive buffer can be empty, it just can't have length zero.

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -149,12 +174,85 @@ impl EAAKbc {
...
+        let trans_data: &[u8] = trans_json.as_bytes();
+
+        let recv_string: String =
+            self.kbs_trans_and_recv(trans_data, DEFAULT_RECV_BYTES_SIZE, "Get Resource Info")?;
+
+        let response = serde_json::from_str::<GetResourceInfoResponse>(recv_string.as_str())?;
+
+        match response.status.as_str() {
+            "OK" => {
+                if let Some(info) = response.data {
```
fitzthum, line 224: maybe `ok_or` and `?` instead of this big `if let`?

src/kbc_modules/eaa_kbc/mod.rs
```diff
@@ -102,14 +118,16 @@ impl EAAKbc {
...
+            self.kbs_trans_and_recv(trans_data, DEFAULT_RECV_BYTES_SIZE, "Version")?;
         let response: VersionResponse =
             serde_json::from_str::<VersionResponse>(recv_string.as_str())?;
 
         match response.status.as_str() {
             "OK" => Ok(response.version),
-            "Fail" => return Err(anyhow!("The VersionResponse status is 'Fail'!")),
-            _ => return Err(anyhow!("Can't understand the VersionResponse status!")),
+            "Fail" => return Err(anyhow!("The VersionResponse status is Fail")),
+            _ => return Err(anyhow!("Can not understand the VersionResponse status")),
```
fitzthum, line 130: nit: `Can not` -> `Cannot`

fitzthum (commented):
Looks good. There are some things I might change about the protocol, but that is out of scope here so I just made a few minor comments.

src/kbc_modules/eaa_kbc/mod.rs
jialez0: The receive here will return the data one by one in the form of slices in order. Because the maximum size of the data transmitted by TLS at one time is limited, each receive in the loop here will return different contents of the data.

sameo (approved):

fitzthum (approved):
LGTM

PR #50, opened 2022-03-23
[sample KBS] Update some codes and docs to keep with the current AA.

jialez0:


sameo (approved):

wainersm:
Hi @jialez0 , a little outside the scope of this PR but while we are here.... could you link `sample_kbs/README.md ` in the "List of supported KBC modules" table in the `README.md`?

jialez0:
@wainersm Done.

> Looks good. At some point we might want to rename `sample_kbs` given that it isn't really a KBS. It's really just a tool for encrypting container images with skopeo.



> There are other places where `grpc_sock` is still mentioned:
> 
> ```
> $ grep -r grpc_sock
> README.md:Start AA and use grpc_sock parameter to specify the endpoint of AA's keyprovider service, e.g, listen on local 47777 port:
> README.md:attestation-agent --grpc_sock 127.0.0.1:47777
> README.md:RUST_LOG=attestation_agent attestation-agent --grpc_sock 127.0.0.1:4777
> sample_kbs/README.md:RUST_LOG=sample_kbs cargo run --features sample_enc --release -- --grpc_sock 127.0.0.1:50000
> sample_kbs/README.md:RUST_LOG=attestation_agent attestation-agent --grpc_sock 127.0.0.1:48888
> sample_kbs/src/enc_mods/offline_fs_kbs/README.md:cargo run --release --features offline_fs_kbs -- --grpc_sock 127.0.0.1:50000
> sample_kbs/src/main.rs:                .long("grpc_sock")
> sample_kbs/src/main.rs:                    for example: --grpc_sock 127.0.0.1:11223",
> src/kbc_modules/eaa_kbc/README.md:RUST_LOG=attestation_agent attestation-agent --grpc_sock 127.0.0.1:50000
> src/kbc_modules/offline_fs_kbc/README.md:attestation-agent --grpc_sock 127.0.0.1:50000
> src/kbc_modules/offline_sev_kbc/README.md:cargo run --release --no-default-features --features offline_sev_kbc -- --grpc_sock 127.0.0.1:50000
> ```

wainersm (approved):

sample_kbs/src/main.rs
```diff
@@ -18,11 +18,11 @@ async fn main() -> Result<()> {
         .version("1.0.0")
         .arg(
             Arg::with_name("socket addr")
-                .long("grpc_sock")
+                .long("keyprovider_sock")
```
jodh-intel, line 21: Since the names are changing, is it too late to consider replacing the underscore with a dash? It looks better imho and is the expected way to specify long options (`--keyprovider-sock`).

Same comment for `--getresource_sock` (rename to `--getresource-sock`).

fitzthum: I think the goal of this PR is to have the the names used by the `sample_kbs` be the same as the names used by the Attestation Agent itself. Maybe we can have a separate issue for whether we want to change the argument name across the entire project?

jialez0: > I think the goal of this PR is to have the the names used by the `sample_kbs` be the same as the names used by the Attestation Agent itself. Maybe we can have a separate issue for whether we want to change the argument name across the entire project?

@jodh-intel @fitzthum Yes, If we decide to change the parameter name, Kata-agent also needs to make changes synchronously. Maybe it's better to open a new PR to solve this problem, if you want to do so.

jodh-intel (approved):

PR #51, opened 2022-03-23
sample_kbs: add instruction to copy image to localhost

wainersm:
On sample_kbs's README tell how to copy an image to the current directory
as it might not be obvious to newcomers.

Signed-off-by: Wainer dos Santos Moschetta <wainersm@redhat.com>

### DIFF
--- a/sample_kbs/README.md
+++ b/sample_kbs/README.md
@@ -29,6 +29,12 @@ vim ocicrypt.conf:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


fitzthum (approved):
LGTM. There are a handful of ways to do this really. You could encrypt while copying from docker to oci, oci to docker, docker to docker, etc. Might as well make the docs more verbose tho.

jialez0 (approved):

PR #54, opened 2022-04-15
Rename sample_kbs to sample_keyprovider

### LINKED ISSUE #241: Rename sample_kbs to sample_keyprovider
The so-called `sample_kbs` is not a KBS. This is an extremely confusing name that has already claimed several victims. Why don't we call it `sample_keyprovider` instead? @jialez0

jialez0:
Fix: confidential-containers/guest-components#241


fitzthum (approved):
LGTM

PR #56, opened 2022-05-16
Extract AA core functions as a library crate.

### LINKED ISSUE #240: Extract AA core functions as a rust library to support enclave-cc
In the architecture of enclave-CC, sometimes only one process can run in enclave (such as Gramine), so the core function of AA needs to be integrated into the agent program as a rust library crate. But the current implementation of AA is a grpc service application. Therefore, if AA want to support enclave-CC, we need to transform the current AA architecture and separate the core function (including KBCs) and the part of grpc service into two rust crates, the former as library crate and the latter as application crate.

I'm not sure about the description of this work in the current development plan of enclave-CC. It may take some time and some tests to complete this work, but it may be a necessary work for AA to support enclave CC.

FYI @sameo @fitzthum  @jodh-intel @jiazhang0

jialez0:
Fix: confidential-containers/guest-components#240


README.md
```diff
@@ -1,23 +1,38 @@
 # Attestation Agent
 
-Attestation Agent (AA for short) is a user space service for attestation procedure. 
-In Kata Confidential Containers (Kata CC for short), AA implements the WrapKey API defined by the key-provider protocol, 
-which is responsible for performing attestation and obtaining the Key Encryption Key (KEK for short) from Key Broker Service (KBS for short),
-or requesting KBS to decrypt the encrypted payload stored in the image layer annotation.
+Attestation Agent (AA for short) is a service function set for attestation procedure
+in Confidential Containers. It provides kinds of service APIs that need to make
+requests to the Rely Party (Key Broker Service) in Confidentail Containers,
```
dcmiddle, line 5:
```suggestion
requests to the Relying Party (Key Broker Service) in Confidential Containers,
```

src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+///     "example_annotation".to_string()
+/// );
+/// ```
+
+/// `ServiceAPIs` defines the service APIs of attestation agent that need to make requests
+///  to the Relying Party (Key Broker Service) in Confidential Containers.
+///
+/// For every service API, the `kbc_name` and `kbs_uri` is necessary, `kbc_name` tells
+/// attestation agent which KBC module it should use and `kbs_uri` specifics the KBS address.
+pub trait ServiceAPIs {
```
fitzthum, line 38: `ServiceAPIs` seems a bit vague. Maybe we could call it `AttestationAPI` or `AgentEndpoints` or something.

src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// The decryption method may be to obtain the key from KBS for decryption, or
+    /// directly send the `annotation` to KBS for decryption, which depends on the
+    /// specific implementation of each KBC module.
+    fn decrypt(&mut self, kbc_name: String, kbs_uri: String, annotation: String)
+        -> Result<Vec<u8>>;
+
+    /// `get_resource` is used to request KBS to obtain confidential resources, including
+    /// confidential data or files.The specific format of `resource_description` is defined by
+    /// different KBC and corresponding KBS.
+    fn get_resource(
```
fitzthum, line 50: `decrypt` and `get_resource` are slightly inconsistent names. First of all, Rust doesn't really use `get_*` so maybe we should change `get_resource` to something else? `decrypt` is also a little bit vague. What are we decrypting?

fitzthum (commented):
A few minor comments. This is basically good. My main concerns are about the name of our traits and methods. Since we are creating a library, now is the time to settle on names that we will stick with.

src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+///     "https://xxxxx".to_string(),
+///     "example_annotation".to_string()
+/// );
+/// ```
+
+/// `ServiceAPIs` defines the service APIs of attestation agent that need to make requests
+///  to the Relying Party (Key Broker Service) in Confidential Containers.
+///
+/// For every service API, the `kbc_name` and `kbs_uri` is necessary, `kbc_name` tells
+/// attestation agent which KBC module it should use and `kbs_uri` specifics the KBS address.
```
dcmiddle, line 37:
```suggestion
/// attestation agent which KBC module it should use and `kbs_uri` specifies the KBS address.
```

src/lib.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// `decrypt`is used to decrypt the encrypted information in `annotation`.
+    /// The specific format of `annotation` is defined by different KBC and corresponding KBS.
+    /// The decryption method may be to obtain the key from KBS for decryption, or
+    /// directly send the `annotation` to KBS for decryption, which depends on the
+    /// specific implementation of each KBC module.
+    fn decrypt(&mut self, kbc_name: String, kbs_uri: String, annotation: String)
+        -> Result<Vec<u8>>;
+
+    /// `get_resource` is used to request KBS to obtain confidential resources, including
+    /// confidential data or files.The specific format of `resource_description` is defined by
```
dcmiddle, line 48:
```suggestion
    /// confidential data or files. The specific format of `resource_description` is defined by
```

README.md
```diff
@@ -1,23 +1,38 @@
...
+The main body of AA is a rust library crate, which contains KBC modules used to communicate
+with various KBS. In addition, this project also provides a gRPC service application, 
+which allows callers to call the services provided by AA through gRPC.
+
+## Library crate
+
+Import AA in `Cargo.toml` of your project with specific KBC(s):
+
+```toml
+attestation-agent = { git = "https://github.com/confidential-containers/attestation-agent", features = ["sample_kbc"] }
```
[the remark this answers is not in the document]
jialez0: That's a good idea, maybe we can mention it in README ？

src/lib.rs
jialez0: What about `download_confidential_resource` and `decrypt_image_layer_anno` ?

dcmiddle (approved):
LGTM!

src/lib.rs
fitzthum: Yeah I think that's good. Might as well say `annotation` instead of `anno`. The name is going to be long either way. You could also just say `decrypt_annotation` or `decrypt_layer_annotation` or something.

fitzthum (approved):

PR #60, opened 2022-06-17
[CC-KBC] Initial implement of CC-KBC.

jialez0:
This PR does not include the functional parts that need to communicate with KBS temporarily, because it needs to wait for [KBS](https://github.com/confidential-containers/kbs) ready before development.


dubek (commented):
Impressive work. Comments below. 

I'm particularly hesitant with our crypto implementation of encrypting/decrypting a message (that combination of RSA and AES), I want to look for a library that does it in one go instead of inventing our own.

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+        Ok(TeePubKey {
+            algorithm: RSA_ALGORITHM.to_string(),
+            pubkey_length: RSA_PUBKEY_LENGTH.to_string(),
+            pubkey: pubkey_pem_string,
+        })
+    }
+
+    // Use TEE private key to decrypt cipher text.
+    pub fn decrypt(&self, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
+        let padding = PaddingScheme::new_pkcs1v15_encrypt();
```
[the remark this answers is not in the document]
jialez0: This can not be a const, but can only be a lazy static. But I don't think it is necessary to use it as a global static variable, because it will only be used here.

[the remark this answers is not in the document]
dubek: Yes. I wanted to keep the RSA_ALGORITHM and the generation of padding close together in the source code (I'm worried for a future change in which we change one but not the other). We can do it using something like:

```rust
const RSA_ALGORITHM: &str = "rsa-pkcs1v15";
const NEW_PADDING: fn() -> PaddingScheme = PaddingScheme::new_pkcs1v15_encrypt;
```

and then in `decrypt()` use:

```rust
let padding = NEW_PADDING();
```

but that's a bit of over-engineering for this.

Maybe some other way of keeping them together? Maybe a CryptoType struct with `name` (string) and `new_padding()` function? Not sure it's worth it.

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    // This symmetric key is used to encrypt API output by KBS.
+    #[serde(rename = "enc-symkey")]
+    enc_symkey: String,
+}
+
+impl CryptoAnno {
+    // Use the TEE private key to unwrap the encrypted symmetric key,
+    // then use the symmetric key to decrypt cipher text.
+    #[allow(unused_assignments)]
+    pub fn decrypt(&self, tee_key: TeeKey, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
```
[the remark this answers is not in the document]
jialez0: We can maintain the current status until we find a good library, and welcome to propose a new PR to optimize it :)

Encryption side: the encryption of symmetric key and output content should take place on the KBS side. Please refer to [this document](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md) for details about the logic here

src/kbc_modules/cc_kbc/kbs_protocol/message.rs
```diff
@@ -0,0 +1,107 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use crate::kbc_modules::cc_kbc::crypto::*;
+use anyhow::*;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+pub const KBS_PROTOCOL_VERSION: &str = "0.1.0";
```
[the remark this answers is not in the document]
jialez0: If this is an integer, we can't accurately describe a version number like "2.1.3".

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+            .ok_or_else(|| anyhow!("Generate TEE key failed"))?;
+        let attester = self
+            .attester
+            .as_ref()
+            .ok_or_else(|| anyhow!("TEE attester missed"))?;
+
+        let tee_pubkey = key
+            .export_pubkey()
+            .map_err(|e| anyhow!("Export TEE pubkey failed: {:?}", e))?;
+        let tee_pubkey_string = serde_json::to_string(&tee_pubkey)?;
```
dubek, line 65: Is the order of the fields in this JSON string identical to what is produced in line 81 (tee_pubkey as part of the Evidence)? Is it defined as a compact JSON string (no spaces, no end-of-line chars)?

We have this in the [evidence json](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestation):

```
	    /* The public key generated by KBC in HW-TEE, it is valid until the next time an attestation is required.
	       Its hash needs to be included in HW-TEE evidence and signed by HW-TEE hardware. */
	    "tee-pubkey": {
		    "algorithm": "",
		    "pubkey-length": "",
		    "pubkey": ""
		},
```

The KBS should be able to serialize this object to compute the hash of nonce+tee_pubkey_string (to verify the report). And the serialization must be identical to how this was done here (in the KBC).

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    // Export TEE public key as specific structure.
+    pub fn export_pubkey(&self) -> Result<TeePubKey> {
+        let pem_line_ending = rsa::pkcs1::LineEnding::default();
+        let pubkey_pem_string = self.public_key.to_public_key_pem(pem_line_ending)?;
+
+        Ok(TeePubKey {
+            algorithm: RSA_ALGORITHM.to_string(),
+            pubkey_length: RSA_PUBKEY_LENGTH.to_string(),
```
dubek, line 46: curious: Why is this value a string? (I saw this is how it is defined in the [protocol doc](https://github.com/confidential-containers/kbs/blob/main/docs/kbs_attestation_protocol.md#attestation))

src/kbc_modules/cc_kbc/kbs_protocol/message.rs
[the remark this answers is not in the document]
dubek: Yes, you're right. I didn't think protocols need a "tree" of versions (where we advance from "2.1.3" both to "2.1.4" and to "2.2.0"), but I see that's what we agreed on in the protocol doc.

src/kbc_modules/cc_kbc/mod.rs
jialez0: Thanks for the reminder. It is really necessary to ensure that the JSON string format is strictly consistent. Otherwise, KBS will not be able to verify the hash correctly. Therefore, we'd better set the tee-pubkey (line 81) contained in evidence to the `String` type, that is, defined as a compact JSON string, right?

src/kbc_modules/cc_kbc/crypto/mod.rs
jialez0: Ah, this may be an oversight when writing KBS protocol documents. It should be more appropriate to use integer types here

src/kbc_modules/cc_kbc/mod.rs
dubek: Option 1 is to set use a string value for `tee-pubkey`, for example:

```
{
  "evidence": {
    "nonce": "",
    "tee": "",
    "tee-pubkey": "{\"algorithm\":\"rsa-pkcs1v15\",\"pubkey-length\":2048,\"pubkey\":\"AABBCCDD...\"}",
    "tee-evidence": {
      ..
    }
  }
}
```

Option 2 is to keep the current format, but the JSON serialization (before hashing) should be done by a canonical-json serializer (both in KBC and KBS). Unfortunately, I don't think serde_json can do this out of the box. I see there's a [cjson crate](https://docs.rs/cjson/latest/cjson/) and a [canonical_json crate](https://docs.rs/canonical_json/latest/canonical_json/) but both look a bit unmaintained. Also there's no clear standard for canonical JSON (or worse: [xkcd 927](https://xkcd.com/927/)).

jialez0: So, option 1 looks better now.

src/kbc_modules/cc_kbc/kbs_protocol/message.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[serde(rename = "extra-params")]
+    pub extra_params: String,
+}
+
+impl Request {
+    pub fn new(token: Option<String>, tee: String) -> Request {
+        Request {
+            version: KBS_PROTOCOL_VERSION.to_string(),
+            tee,
+            token: token.unwrap_or_else(|| "".to_string()),
```
dubek, line 30: Simpler:

    token: token.unwrap_or("".to_string()),

or even:

    token: token.unwrap_or_default(),

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    fn decrypt_payload(&mut self, _annotation: &str) -> Result<Vec<u8>> {
+        Err(anyhow!("Decrypt Payload API of this KBC is unimplemented."))
+    }
+}
+
+impl Kbc {
+    pub fn new(_kbs_uri: String) -> Kbc {
+        // Detect TEE type of the current platform.
+        let tee_type = detect_tee_type().ok().unwrap_or(TEE::Unknown);
```
dubek, line 38: Why do we call `ok()` here? I think you can call `unwrap_or()` directly on the `Result<TEE>`.

(or, modify `detect_tee_type` above to return `TEE::Unknown` instead of Err(...). And then there's no need for result/option. )

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        // Create attester instance.
+        let attester = tee_type.to_attester().ok();
```
dubek, line 41: Errors in `detect_tee_type` or `to_attester` are swallowed and not reported anywhere. Shouldn't we completely fail in these cases? Why wait for `generate_evidence` to fail?

Or, thinking about this again: currently the Kbc struct has 3 Option fields. Can the Kbc work correctly when they are set to None? If not, they should not be Option; this way we won't construct a Kbc struct that will fail on every usage.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -0,0 +1,93 @@
...
+
+        Ok(Evidence {
+            nonce: self.nonce.clone(),
+            tee: self.tee.clone(),
+            tee_pubkey: tee_pubkey_string,
+            tee_evidence,
+        })
+    }
+
+    fn dump_plain_output(&self, response: Response) -> Result<Vec<u8>> {
```
dubek, line 86: For me, `dump` usually means that it goes to stdout or file (or to the garbage).

Here I think that a clearer name may be `decrypt_response_output` ? Or `decrypt_kbs_response` ?

dubek (approved):
Looks good, but see some small comments below:

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+        Ok(TeePubKey {
+            algorithm: RSA_ALGORITHM.to_string(),
+            pubkey_length: RSA_PUBKEY_LENGTH,
+            pubkey: pubkey_pem_string,
+        })
+    }
+
+    // Use TEE private key to decrypt cipher text.
+    pub fn decrypt(&self, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
+        let padding = NEW_PADDING();
```
fitzthum, line 65: I'm a little confused about this const fn. Why is it needed?

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+// The key inside TEE to decrypt confidential data.
+#[derive(Debug, Clone)]
+pub struct TeeKey {
+    private_key: RsaPrivateKey,
+    public_key: RsaPublicKey,
+}
+
+// The public key of TEE.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct TeePubKey {
```
fitzthum, line 29: So this is just the struct that we use to export the pubkey, right? It might be worth mentioning that in the comment since we have another (different) pubkey below.

src/kbc_modules/cc_kbc/kbs_protocol/message.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Response {
+    // The API output of KBS.
+    // It is encrypted and base64 encoded by the random symmetric key generated by KBS.
+    output: String,
+
+    // Crypto annotation record the information of the symmetric key to encrypt output.
+    #[serde(rename = "crypto-annotation")]
+    crypto_annotation: CryptoAnnotation,
```
fitzthum, line 71: In the KBS repo, there is a separate `enc-symkey` field in the response. Here it is part of the CryptoAnnotation. Might be worth updating the doc.

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,130 @@
...
+    // The symmetric key encrypted by TEE's public key.
+    // This symmetric key is used to encrypt API output by KBS.
+    // Base64 encoded.
+    #[serde(rename = "enc-symkey")]
+    enc_symkey: String,
+}
+
+impl CryptoAnnotation {
+    // Use the TEE private key to unwrap the encrypted symmetric key,
+    // then use the symmetric key to decrypt cipher text.
```
fitzthum, line 91: Is the symmetric key going to be randomized for each response? Otherwise it seems like we could just decrypt it once and store it in the KBC alongside the token.

fitzthum (approved):
A few minor things, but I think we can keep moving with the rest of the KBC stuff.

src/kbc_modules/cc_kbc/crypto/mod.rs
dubek: The idea was to keep the definitions of `RSA_ALGORITHM` and `NEW_PADDING` next to each other, because they are coupled (should be changed together, if we change them in the future). They are at the top of this file:

```rust
const RSA_ALGORITHM: &str = "rsa-pkcs1v15";
const RSA_PUBKEY_LENGTH: usize = 2048;
const NEW_PADDING: fn() -> PaddingScheme = PaddingScheme::new_pkcs1v15_encrypt;
```

I agree that the solution (which I suggested) is not very elegant or "rusty" ...

Suggestions welcome! Maybe the original was better...

src/kbc_modules/cc_kbc/mod.rs
jialez0: Now according to the [KBC framework](https://github.com/confidential-containers/attestation-agent/blob/main/docs/kbc_module_development_guide.md), `fn new(_kbs_uri: String) -> Kbc` can not return a Result, so I have to temporarily use option in the KBC structure to surround the err returned by `Teekey::new()` and `to_attester()`. Maybe we can change the return value of the KBC's `new()` function later, but this will cause all KBCs in AA need to be modified.

dubek: Yes, you're right. Thanks for pointing that out.

I also see that the KBC framework instantiates the kbc modules lazily, upon first request (for example in [`decrypt_image_layer_annotation()`](https://github.com/confidential-containers/attestation-agent/blob/main/src/lib.rs#L113)). Even if we change the return type of `new()` to `Result<Kbc>`, we'll encounter the error quite encounter the error quite late (and not an early fail when the attestation-agent starts, which is what I was trying to achieve).

I hope we can improve it later (in the framework level); but not in this PR.

src/kbc_modules/cc_kbc/crypto/mod.rs
jialez0: Whether the symmetric key is randomized for each response depends on the security policy adopted by KBS, but I think maybe it is not a good idea to save this key in KBC, because it will cause us to design additional communication fields to update this key (for example, when KBS finds that there is a risk of key disclosure).

src/kbc_modules/cc_kbc/crypto/mod.rs
jialez0: @fitzthum This is a necessary work in the standard usage of `rsa` crate. For details, you can refer to [this](https://docs.rs/rsa/0.6.1/rsa/index.html).

src/kbc_modules/cc_kbc/kbs_protocol/message.rs
jialez0: This field is actually in `crypto-annotation`, but it looks like an separate field in the KBS document due to the confusion of JSON indentation. I have corrected the confusion of locking in this [pr](https://github.com/confidential-containers/kbs/pull/2).

src/kbc_modules/cc_kbc/crypto/mod.rs
fitzthum: Ok, makes sense. I've only seen people use const functions when they want to use a function to define another const, but this seems reasonable.

src/kbc_modules/cc_kbc/crypto/mod.rs
fitzthum: Ok. Makes sense to leave it up to the KBS. Is there anything that prevents replay of a response?

jialez0: The replay attack defense of response is the responsibility of the protocol that transmits these messages (HTTPS). These JSON structures defined by us will be passed as HTTPS payload.

jialez0: In order to avoid maliciously impersonating KBS, KBS's TLS certificate need to be preset in the environment in which KBC runs, but I'm not sure how to achieve this currently...

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+            AES_GCM_256_ALGORITHM => {
+                let decrypting_key = aes_gcm::Key::from_slice(&symkey);
+                let aes_gcm_cipher = Aes256Gcm::new(decrypting_key);
+
+                let iv_decoded = base64::decode(self.initialization_vector.clone())?;
+                let nonce = aes_gcm::Nonce::from_slice(&iv_decoded);
+
+                aes_gcm_cipher
+                    .decrypt(nonce, cipher_text.as_ref())
+                    .map_err(|e| anyhow!("AES GCM decrypt failed: {:?}", e))?
```
jodh-intel, line 109: There seem to be multiple varieties of the AES GCM algorithm so why not just log `AES_GCM_256_ALGORITHM` as that is terse but fully descriptive?

src/kbc_modules/cc_kbc/attester/sample/mod.rs
```diff
@@ -0,0 +1,40 @@
...
+
+use super::Attester;
+use anyhow::*;
+use serde::{Deserialize, Serialize};
+use std::env;
+
+// If the environment variable "AA_SAMPLE_ATTESTER_TEST" is set,
+// the TEE platform is considered as "sample".
+pub fn detect_platform() -> bool {
+    env::var("AA_SAMPLE_ATTESTER_TEST").is_ok()
```
jodh-intel, line 14: If setting this env-var causes the behaviour of this component to become a NOP, that worries me: could this be a way to compromise the security of the system?

Also, the name of this variable is a bit ambiguous as "_TEST" might imply this functionality is simply added for testing. Is that the case, or is this just the variable used "to test" if we're running with a sample TEE?

jodh-intel (commented):
Thanks @jialez0 - a few comments...

src/kbc_modules/cc_kbc/crypto/mod.rs
fitzthum: Yeah we have a bit of a chicken-and-egg problem here. It seems like we want the KBS identity to be measured somehow, but since the KBC is the one that validates the measurement this might not actually give us the guarantee that we want.

One potential answer is that we need to securely expose the KBS identity to the guest owner directly or to other services operating on behalf of the guest owner. In other words, we need to extend the chain of trust from the HW to the KBS to some other entity like a database providing secure data.

This is kind of a complicated problem that we probably don't need to worry about here yet. I am planning to make another issue about it soon.

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+        })
+    }
+
+    // Use TEE private key to decrypt cipher text.
+    pub fn decrypt(&self, cipher_text: Vec<u8>) -> Result<Vec<u8>> {
+        let padding = NEW_PADDING();
+
+        self.private_key
+            .decrypt(padding, &cipher_text)
+            .map_err(|e| anyhow!("TEE rsa key decrypt failed: {:?}", e))
```
jodh-intel, line 69:
```suggestion
            .map_err(|e| anyhow!("TEE RSA key decrypt failed: {:?}", e))
```

sameo (approved):

src/kbc_modules/cc_kbc/crypto/mod.rs
jialez0: I'm not sure whether other algorithms added in the future will need padding (at least RSA needs it at present), so maybe we can consider code optimization here when new algorithms are added?

fitzthum:
Let's try to get this merged. @jodh-intel, any further comments? @hdxia Any comments form the KBS perspective?

src/kbc_modules/cc_kbc/crypto/mod.rs
jodh-intel: @jialez0 - Yep - it was just an idea. And we could of course make the handler optional:

```rust
handler: Option<fn() -> PaddingScheme>,
```

jodh-intel (approved):
Thanks @jialez0 - lgtm. We can iterate it further as needed in subsequent PRs.

PR #64, opened 2022-07-04
build: support building attestation agent for s390x

mattarnoatibm:
Support building attestation agent for s390x via cross-compilation
on x86_64 (Debian-based) os.

Support native build for attestation agent for s390x.

Links to issues

* https://github.com/confidential-containers/guest-components/issues/237
* https://github.com/confidential-containers/guest-components/issues/236

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -32,6 +32,10 @@ jobs:
       - name: Musl build with default features
         run: |
           make LIBC=musl
+      
+      - name: s390x build with default features
+        run:
+          make ARCH=s390x
 
       - name: Run cargo test
         uses: actions-rs/cargo@v1
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -33,9 +33,9 @@ jobs:
         run: |
           make LIBC=musl
       
-      - name: s390x build with default features
+      - name: s390x build with offline_fs_kbc feature
         run:
-          make ARCH=s390x
+          make ARCH=s390x KBC=offline_fs_kbc
 
       - name: Run cargo test
         uses: actions-rs/cargo@v1
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,12 @@ ifeq ($(KBC), eaa_kbc)
     RUSTFLAGS_ARGS += -C link-args=-Wl,-rpath,/usr/local/lib/rats-tls
 endif
 
+ifeq ($(KBC), offline_sev_kbc)
+    ifeq ($(ARCH), s390x)
+        $(error ERROR: Offline SEV KBC does not support s390x architecture!)
+    endif
+endif
+
 ifneq ($(RUSTFLAGS_ARGS),)
     RUST_FLAGS := RUSTFLAGS="$(RUSTFLAGS_ARGS)"
 endif
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,6 @@ KBC ?=
 DESTDIR ?= $(PREFIX)/bin
 RUSTFLAGS_ARGS ?=
 
-PROTOC_S390X_VERSION := 21.1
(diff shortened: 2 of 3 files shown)


dubek:
> Is it sensible to not support a s390x build for the attestation agent with the offline_sev_kbc feature

Yes, IMO that makes sense. Even though the code of offline_sev_kbc doesn't have any AMD-specific instructions, it relies on the `efi_secret` kernel module, which itself is only built for x86_64 Linux kernels.

.github/workflows/basic.yml
```diff
@@ -32,6 +32,10 @@ jobs:
       - name: Musl build with default features
         run: |
           make LIBC=musl
+      
+      - name: s390x build with default features
+        run:
+          make ARCH=s390x
```
dubek, line 38: don't you want to build also with `offline_fs_kbc` which is actually used for CC in s390x Secure Execution?

mattarnoatibm:
> > Is it sensible to not support a s390x build for the attestation agent with the offline_sev_kbc feature
> 
> Yes, IMO that makes sense. Even though the code of offline_sev_kbc doesn't have any AMD-specific instructions, it relies on the `efi_secret` kernel module, which itself is only built for x86_64 Linux kernels.

Thanks. I updated the Makefile so that building using `make` when `KBC=offline_sev_kbc` and the target architecture is s390x throws an error saying  the offline SEV KBC does not support the s390x architecture.

Makefile
```diff
@@ -9,24 +9,52 @@ DEBIANOS := $(shell cat /etc/debian_version 2> /dev/null)
...
+        RUST_TARGET_FOR_TARGET_ARCH_INSTALL := $(shell rustup target add ${RUSTC_TARGET_FOR_TARGET_ARCH})
+        RUSTFLAGS_ARGS += -C linker=$(GCC_COMPILER_FOR_TARGET_ARCH)
+    else
+        $(error ERROR: Cross-compiling is only tested on Debian-like OSes)
+    endif
+endif
+
+ifeq ($(SOURCE_ARCH), s390x)
+    PROTOC_BINARY_INSTALL := $(shell wget https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_S390X_VERSION}/${PROTOC_S390X_ARCHIVE} && unzip -u ${PROTOC_S390X_ARCHIVE} && sudo cp bin/protoc /usr/local/bin/ && rm -f ${PROTOC_S390X_ARCHIVE})
+endif
```
dubek, line 53: Isn't this some kind of prerequisite?
How do we know that protoc is installed for in x86_64 machines?
Maybe this should be outside the Makefile?
(similar question for `apt-get install ...` stuff we do here)

mattarnoatibm: Installing the separate `protoc` binary is required to compile attestation agent on s390x.

On x86_64 machines the attestation agent compiles without needing `protoc`.

I'm not sure if it's better to put the install for the protoc dependency and the `apt-get install ...` for the gcc cross-compiler toolchain inside the Makefile or outside of it. When building for attestation agent for x86_64 with the musl toolchain the installation of the `musl-tools` package is done inside the Makefile.

I don't maintain the build with musl but maybe that the musl-tools package is installed inside the Makefile means there's a precedent for putting installation of target-specific dependencies inside the Makefile.

dubek: > I don't maintain the build with musl but maybe that the musl-tools package is installed inside the Makefile means there's a precedent for putting installation of target-specific dependencies inside the Makefile.

Yes, you're right. OK.

stevenhorsman:
@jialez0 - Hey Jiale - are able to take a look at this change as I think you've done most of the Makefile updates? Thanks!

sameo (approved):

jodh-intel (approved):
Thanks @mattarnoatibm.

lgtm

PR #66, opened 2022-07-25
Make AA and KBC traits async

fitzthum:
The upcoming `online_sev_kbc` and `CC-KBC` will need async support. This should enable it. 

I think that we should probably rework our usage of the mutex in the gRPC functions at some point. For now I'm just leaving it as it is. This isn't really a performance-sensitive application, but I think it's still non-optimal.

The `eaa-kbc` seems a little bit broken right now. As you can see the test has some problem. There are also a few clippy warnings that should be fixed at some point.

@dubek

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,10 @@ strum = { version = "0.24.0", features = ["derive"] }
 rsa = "0.6.1"
 sha2 = "0.10"
 rand = "0.8.5"
+async-trait = "0.1.56"
+
+[dev-dependencies]
+tokio = { version = "1.20.1", features = ["macros", "rt-multi-thread"] }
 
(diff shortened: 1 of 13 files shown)


dubek (approved):
Looks simple and good to me.

fitzthum:
I have an `sev_online_kbc` implemented on top of this PR. I will post once we merge this.

fitzthum:
I've re-run the tests after https://github.com/confidential-containers/attestation-agent/pull/67 but the EAA-KBC test still isn't happy?

jialez0:
@fitzthum Maybe you can try rebase you branch to upstream, and then force push. I think the problem will be solved :)

jialez0 (approved):
LGTM.

PR #67, opened 2022-07-26
Fix EAA KBC CI failed

jialez0:


.github/workflows/eaa_kbc.yml
```diff
@@ -14,45 +16,35 @@ jobs:
...
     name: Check
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
         rust:
           - stable
+
+    container: runetest/compilation-testing:ubuntu18.04
```
fitzthum, line 26: Is there a 20.04 version of this container? Why are we using the `rune` container here?

fitzthum (approved):
This looks much nicer. I just have one question about container image that we are using.

.github/workflows/eaa_kbc.yml
```diff
@@ -14,45 +16,35 @@ jobs:
...
-
-      - name: Create rats-tls compiling container
+      - name: Install Rust toolchain
         run: |
-          rats_compile_env=$(docker run -itd --privileged --rm --net host -v /usr/local/lib:/usr/local/lib inclavarecontainers/test:compile-check-ubuntu18.04);
-          echo "rats_compile_env=$rats_compile_env" >> $GITHUB_ENV
+          apt-get update && apt-get install -y cargo;
+          cp -r /root/.cargo /github/home/.cargo;
+          rustup install stable;
+          rustup default stable;
```
dubek, line 39: 1. What was wrong with the rust installation via `actions-rs/tollchain` ?
2. Doesn't rustup install cargo? Do we need to install cargo via apt?
3. I think the semicolons at the end of the lines are not needed in this file (everywhere). But please double-check.

.github/workflows/eaa_kbc.yml
```diff
@@ -14,45 +16,35 @@ jobs:
...
 
       - name: Build and install rats-tls
         run: | 
-          docker exec $rats_compile_env bash -c 'git clone https://github.com/alibaba/inclavare-containers;
-          cd inclavare-containers/rats-tls;
+          git clone https://github.com/inclavare-containers/rats-tls;
+          cd rats-tls;
+          apt-get install -y libcurl4-openssl-dev;
```
dubek, line 45: I suggest moving the `apt-get install` line to be the first in this section.

dubek (approved):
Generally looks good (especially if it solves the CI issue...).

Couple of questions (mainly style) about these changes below:

.github/workflows/eaa_kbc.yml
jialez0: `rune` is a historical name. Because the rats-tls that EAA KBC relies on was originally a subproject under the `rune` project, its compilation test container is placed under the runetest namespace. I learned that 20.04 version of this container is under development.

Since this is just a compilation test, we can temporarily use the old version of the container to compile rats-tls.

.github/workflows/eaa_kbc.yml
jialez0: I tried to use `actions-rs/tollchain`, but there will be some problems in the container, because the default installation path is /`root/.cargo`, and `actions-rs/tollchain` will be installed to `/home/github/.cargo`. 

`apt-get install cargo` will install rustup and cargo at the same time, which is simpler than installing rustup first and then installing cargo with rustup.

dubek (approved):
👍

PR #69, opened 2022-08-01
Sample KBC: support get_resource interface for test.

jialez0:
Support get_resource interface in sample KBC for CI test of CC. @arronwy


src/kbc_modules/sample_kbc/mod.rs
```diff
@@ -47,6 +47,20 @@ impl KbcInterface for SampleKbc {
...
+    fn get_resource(&mut self, description: String) -> Result<Vec<u8>> {
+        let desc: ResourceDescription =
+            serde_json::from_str::<ResourceDescription>(description.as_str())?;
+
+        match desc.name.as_str() {
+            "Policy" => Ok(std::include_str!("policy.json").as_bytes().to_vec()),
+            "Sigstore Config" => Ok(std::include_str!("sigstore_config.yaml")
+                .as_bytes()
+                .to_vec()),
+            "GPG Keyring" => Ok(std::include_str!("pubkey.gpg").as_bytes().to_vec()),
```
fitzthum, line 60: Is there any way we can avoid hard coding the name of each resource inside of every KBC? Maybe we should have a shared enum for this?

fitzthum (approved):
Looks fine for a sample, but see my comment about the resource names.

btw @stevenhorsman  this is roughly the code that you might want to add to `offline_fs_kbc` (just read the resources from the fs rather than using the include_str thing).

src/kbc_modules/sample_kbc/mod.rs
jialez0: Done.

PR #70, opened 2022-08-02
Add online_sev_kbc

fitzthum:
Introduction of `online_sev_kbc`.

There is a lot of code overlap between `online_sev_kbc`, `offline_sev_kbc`, and `offine_fs_kbc`. I have a plan to merge these together, but I want to get the KBC working first. 

This PR only supports the keyprovider interface, but I will introduce support for get_resource shortly.

See https://github.com/confidential-containers/simple-kbs/pull/32 for the server side.

@dubek


fitzthum:
I'll fix-up lint issues in the morning.

src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+        let request = tonic::Request::new(OnlineSecretRequest {
+            client_id: connection.client_id.as_hyphenated().to_string(),
+            secret_requests: vec![secret_request],
+        });
+
+        let response = client.get_online_secret(request).await?.into_inner();
+
+        let iv_bytes = base64::decode(response.iv)?;
+        let payload_bytes = base64::decode(response.payload)?;
+        let key_bytes = base64::decode(connection.key.clone())?;
```
dubek, line 164: (future) It seems that all our crypto/keys code (in AA, AS, KBSs) would gain from marking such variables as `Secret` using the [secrecy crate](https://docs.rs/secrecy/latest/secrecy/). That's for a separate discussion/PR.

src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+        OnlineSevKbc {
+            kbs_info: HashMap::new(),
+            ciphers: get_ciphers(),
+            kbs_uri,
+            connection: load_connection(),
+        }
+    }
+
+    async fn query_kbs(&self, secret_type: String, secret_id: String) -> Result<Vec<u8>> {
+        let uri = format!("http://{}", self.kbs_uri).parse::<Uri>()?;
```
dubek, line 138: I think we already mentioned (elsewhere) that this is confusing -- a URI should have a scheme such (as `http://`) at the beginning.

src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+    ciphers: Ciphers,
+    kbs_uri: String,
+    connection: Result<Connection>,
+}
+
+fn get_ciphers() -> Ciphers {
+    // The sample KBC uses aes-gcm (Rust implementation). The offline file system KBC uses OpenSSL
+    // instead to get access to hardware acceleration on more platforms (e.g. s390x). As opposed
+    // to aes-gcm, OpenSSL will only allow GCM when using AEAD. Because authentication is not
+    // handled here, AEAD cannot be used, therefore, CTR is used instead.
```
dubek, line 91: I don't understand this.

I see we do use `Aes256Gcm` to decrypt the blob in the `OnlineSecretResponse`.

And we use openssl `aes_256_ctr()` in decrypt_payload. I assume this needs to support the different ciphers in which those payloads are encrypted (is the decryption request coming from from ocicrypt?).

src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+            kbs_uri,
+            connection: load_connection(),
+        }
+    }
+
+    async fn query_kbs(&self, secret_type: String, secret_id: String) -> Result<Vec<u8>> {
+        let uri = format!("http://{}", self.kbs_uri).parse::<Uri>()?;
+
+        let channel = tonic::transport::Channel::builder(uri).connect_lazy();
+        let mut client = KeyBrokerServiceClient::new(channel);
```
dubek, line 141: (future) Consider setting-up channel+client in `new()` and re-using them in each request, this might improve performance if you need to fetch several secrets in a short time. Though we'll need to see how this works when parallel requests are made.

src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -0,0 +1,209 @@
...
+
+const KEYS_PATH: &str = "/sys/kernel/security/coco/efi_secret/1ee27366-0c87-43a6-af48-28543eaf7cb0";
+const SECRET_MODULE_NAME: &str = "efi_secret";
+const MODPROBE_PATH: &str = "/usr/sbin/modprobe";
+const MOUNT_PATH: &str = "/usr/bin/mount";
+
+type Ciphers = HashMap<String, Cipher>;
+
+#[derive(Deserialize)]
+pub struct Connection {
```
dubek, line 36: why `pub` ?
also re-consider for other definitions in this file.

src/kbc_modules/online_sev_kbc/mod.rs
fitzthum: Yeah so this comment comes originally from the `offline_fs_kbc` implementation. Jakob wanted to use OpenSSL for that because it has better s390x support, but I guess OpenSSL has a quirk where you can only use `aes-gcm` if you have AEAD, which we don't have. As a result he switched to CTR and wrote this comment to explain why he isn't used the same scheme as the sample kbs. For our purposes I don't think this matters much. 

Decrypt payload doesn't deal with the actual container image layers. It just decrypts annotation that contains the actual layer key. So the encryption scheme used here is independent of the image and even of ocicrypt. The encryption scheme for the annotation is really set when we encrypt the container. In this case we have the sample-keyprovider tool which has an offline fs module in it. The annotations are designed to support different encryption schemes but so far that isn't really implemented, which isn't great for our goal of running off-the-shelf encrypted containers. 

Now finally there might be a question of why we are using OpenSSL for the annotations but not for the blobs from the KBS. This is just a quirk of implementation. It might be nicer to converge these although we might run into the same issue with `aes-gcm`.

src/kbc_modules/online_sev_kbc/README.md
```diff
@@ -0,0 +1,24 @@
...
+`simple-kbs` generates a new symmetric key for every connection. The KBC requests each online secret with a randomized guid to prevent replay.
+
+## Usage
+
+This KBC expects a `KBS_URI` parameter. The KBC will not function without the EFI Secret module. The module should be available but not loaded before KBC is invoked. The KBC will not be able to unload the module if /proc has not been mounted.
+
+To run:
+
+```
+make KBS=offline_sev_kbc && make install
```
dubek, line 22: s/offline/online/

dubek (approved):
Looks good to me.

Later we can extract the common functions (with other KBCs) to some kind of utils sub-crate.

Small typo below:

jialez0 (approved):
LGTM.

PR #72, opened 2022-08-12
Tweaks for first release with 5.19 kernel

fitzthum:
In preparation for the first release, which will use the upstream 5.19 kernel and a Ubuntu 20.04 rootfs for SEV, we need a few minor changes.

@dubek

### DIFF
--- a/src/kbc_modules/offline_sev_kbc/mod.rs
+++ b/src/kbc_modules/offline_sev_kbc/mod.rs
@@ -14,7 +14,7 @@ use std::collections::HashMap;
 use std::fs;
 use std::process::Command;
 
-const KEYS_PATH: &str = "/sys/kernel/security/coco/efi_secret/e6f5a162-d67f-4750-a67c-5d065f2a9910";
+const KEYS_PATH: &str = "/sys/kernel/security/secrets/coco/e6f5a162-d67f-4750-a67c-5d065f2a9910";
 const SECRET_MODULE_NAME: &str = "efi_secret";
 const MODPROBE_PATH: &str = "/usr/sbin/modprobe";
 const MOUNT_PATH: &str = "/usr/bin/mount";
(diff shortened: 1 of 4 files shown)


dubek (approved):
LGTM.

I see that in kata agent they use `/sbin/modprobe` as well, so the change here is good.
Maybe we should sometime replace executing `MOUNT_PATH` with `nix::mount::mount` (which is used in kata agent).

src/kbc_modules/offline_sev_kbc/mod.rs
```diff
@@ -14,10 +14,10 @@ use std::collections::HashMap;
 use std::fs;
 use std::process::Command;
 
-const KEYS_PATH: &str = "/sys/kernel/security/coco/efi_secret/e6f5a162-d67f-4750-a67c-5d065f2a9910";
+const KEYS_PATH: &str = "/sys/kernel/security/secrets/coco/e6f5a162-d67f-4750-a67c-5d065f2a9910";
```
jodh-intel, line 17: How about also updating `SecretKernelModule::new()` so that the existing message (`"Failed to load secret module."`) is enhanced to mention that the user needs a 5.19+ kernel?

jodh-intel (commented):
Thanks @fitzthum. One comment...

src/kbc_modules/offline_sev_kbc/mod.rs
dubek: But such a message might not be accurate -- we hope that distributions will backport the efi_secret module to the kernels they are shipping, which might be older than 5.19.

jialez0 (approved):
@fitzthum You can rebase this PR and the CI failure will be resolved.

src/kbc_modules/offline_sev_kbc/mod.rs
fitzthum: I'll make sure that the README says something about 5.19, but there are a few corners where an updated message might be confusing.

PR #75, opened 2022-08-17
Sample KBC: delete unused field scheme

Xynnn007:
Due to https://github.com/confidential-containers/image-rs/issues/40
in image-rs, the field `scheme` specifying
the signing scheme is abandoned now.

This issue here will block https://github.com/confidential-containers/image-rs/issues/43


jialez0 (approved):
LGTM.

PR #78, opened 2022-08-23
Offline file system KBC: support get_resource interface for test

### LINKED ISSUE #235: offline_fs_kbc: Add support for get_resource API to enable signature verification with local attestation
**As a** user of the attestation-agent with the offline_fs_kbc module
**I want to** be able to get the policy, sigstore config and public GPG details
**so that** I can use them for signature verification

### Description

In order to remove the dependency on Confidential containers `skopeo` for our [local/offline attestation scenario](https://docs.google.com/presentation/d/1vVN97q3bosJbjyco4dd_25PfnIYNaK3Kg_wCAmvmS4c/edit#slide=id.g132c3a0b3c7_0_10), one of the things we want, is for the `offline_fs_kbc` KBC to support signature verification through `image-rs` (see [#kata-containers/kata-containers4581](https://github.com/kata-containers/kata-containers/issues/4581) for more information).

To do this I believe the Offline filesystem KBC needs to handle the `Policy`, `Sigstore Config` and `GPG keyring` getResource requests. At the moment it only seems to support the [`decrypt_payload`](https://github.com/confidential-containers/attestation-agent/blob/main/src/kbc_modules/offline_fs_kbc/mod.rs#L33) API, so we want to extend it to support the [`get_resource` interface](https://github.com/confidential-containers/attestation-agent/blob/dfb0f1e85c5f498f11b28c71394d68eec8f013a6/src/kbc_modules/mod.rs#L26) as well, like the `eaa_kbc` does. I'm guessing that this requirement is also applicable to the `offline_sev_kbc` as well, so that might need an equivalent issue.

In absence of a better design (which might exist and I just can't find), we could add a new file, called something like `aa-offline_fs_kbc-resources.json` (as a companion to [`aa-offline_fs_kbc-keys.json`](https://github.com/confidential-containers/attestation-agent/blob/main/sample_keyprovider/src/enc_mods/offline_fs_kbs/README.md#usage)) and read the requirement fields from there based on their field names.
We might want to consider if a single json file should support multiple public gpg keys, but I'm not sure if the `image-rs` code base that we want to integrate with handles these. I'd be interested to know how EAA/verdictd are configured and if there is anything we could learn from them.

### Update

Since I originally wrote this issue, similar function has been written for the sample_kbc, so using confidential-containers/attestation-agent#69 as a base is probably a good way to start and the code not too different to what I'd expect we need.

liudalibj:
Fixes https://github.com/confidential-containers/guest-components/issues/235

- Add get_resource interface in offline_fs_kbc for test
Signed-off by: Da Li Liu <liudali@cn.ibm.com>


src/kbc_modules/offline_fs_kbc/README.md
```diff
@@ -18,6 +18,21 @@ Keys must be provided in the guest file system at `/etc/aa-offline_fs_kbc-keys.j
...
+}
```
Xynnn007, line 27: I'm not sure whether the `kbs://` url format in https://github.com/confidential-containers/image-rs/issues/50 is going to be used in future. And if so, think about the 2 extra optional fields `<sub-type>` and `<tag>`. I'm not quite sure whether it is easy to handle them with this one-layer key-value structure without hard-code.

But it's not important for now, I think, we can handle this potential issue when it comes to have new url format.

src/kbc_modules/offline_fs_kbc/mod.rs
```diff
@@ -53,6 +60,27 @@ impl KbcInterface for OfflineFsKbc {
...
         decrypt(*cipher, key, Some(&iv), &wrapped_data).map_err(|_| anyhow!("Failed to decrypt"))
     }
+
+    async fn get_resource(&mut self, description: String) -> Result<Vec<u8>> {
+        let desc: ResourceDescription =
+            serde_json::from_str::<ResourceDescription>(description.as_str())?;
+        let resources = self.resources.as_ref().map_err(|e| anyhow!("{}", e))?;
+        match ResourceName::from_str(desc.name.as_str()) {
+            Result::Ok(ResourceName::Policy) => Ok(resources
+                .get(POLICY_KEY_NAME)
```
Xynnn007, line 70: The same as mentioned. What if recored this `ResourceName` instead of a file name as a key in the `aa-offline_fs_kbc-resources.json`? A `Resources` object can be instantiated by one parse and then serve this function, and no intermediate constants like `POLICY_KEY_NAME` here would be recorded. I'm not quite sure about this. How do you think of this?

src/kbc_modules/offline_fs_kbc/README.md
liudalibj: Yeah, agree with you, this pr is only an initial pr, we can handle the new `kbs://` URI format in future PR.

src/kbc_modules/offline_fs_kbc/mod.rs
liudalibj: Yeah, agree with you, this pr is only an initial pr, use simple hardcode file name as key name and only support one public key. 
We will add the multi public gpg key support in future PR, maybe replace the constants like `POLICY_KEY_NAME` to related `ResourceName`.
Eg if we choose use `kbs://<sub-type>/<name>:<tag>` schema.

request a default policy maybe the name looks like: `kbs://policy/default:default`
request another policy maybe the name looks looks like: `kbs://policy/vendor:tag1`
request a public gpg key maybe the name looks like: `kbs://pubkey/vendor:tag1`
request another public gpg key maybe the name looks like: `kbs://pubkey/review:tag2`

the local json file looks like:
{ 
  "policy/default:default": base64-encoded-strings,
  "policy/vendor:tag1": base64-encoded-strings,
  "pubkey/vendor:tag1": base64-encoded-strings
  "pubkey/review:tag1": base64-encoded-strings
  "pubkey/review:tag2": base64-encoded-strings
  "pubkey/reviewer2:tag2": base64-encoded-strings
}

liudalibj: remove the hardcode key mapping.
use  `ResourceName` as the key in local json file, so that we can support get multi  resources without codes change.

src/kbc_modules/offline_fs_kbc/README.md
liudalibj: @Xynnn007 I updated the codes and document use the `ResourceName` as the key, so that when can support `kbs://<sub-type>/<name>:<tag>` without codes change in future, please help to check again.

Xynnn007: Thanks, I think it now is better!

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -50,9 +51,23 @@ pub fn load_keys(keyfile_name: &str) -> Result<Keys> {
...
+            Ok(resource) => Ok((k.clone(), resource)),
+            Err(_) => Err(anyhow!("Failed to decode resource")),
+        })
+        .collect()
+}
+
 pub mod tests {
     pub use super::*;
-
+    use crate::kbc_modules::ResourceName::*;
```
fitzthum, line 70: I would change this to `use crate::kbc_modules::ResourceName;` and then use `ResourceName::foo` below. I think that would be a bit clearer.

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -50,9 +51,23 @@ pub fn load_keys(keyfile_name: &str) -> Result<Keys> {
...
     pub use base64::encode;
```
fitzthum, line 71: Similarly I would just import `base64` here.

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -82,6 +102,31 @@ pub mod tests {
         keyfile_path
     }
 
+    #[allow(dead_code)]
```
fitzthum, line 105: I think you can avoid this dead code warning if you move this function out of the test module and then import it.

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -82,6 +102,31 @@ pub mod tests {
...
+    pub fn create_resources_file(name: &str) -> PathBuf {
```
fitzthum, line 106: This should return a result. I think it might also make more sense if it takes a `PathBuf` rather than returning one.

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -82,6 +102,31 @@ pub mod tests {
...
+        let temp_dir = env::temp_dir();
+        let resources_file_path = Path::new(&temp_dir).join(name);
+
+        fs::write(
+            resources_file_path.clone(),
+            format!(
```
fitzthum, line 112: You could probably just define this entire thing as a constant and skip having to format it. There are also nicer ways to make json; either with the `json!` macro or by serializing a struct.

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -62,7 +77,12 @@ pub mod tests {
     pub const KID: &str = "foo";
     #[allow(dead_code)]
     pub const KEY: [u8; 32] = *b"passphrasewhichneedstobe32bytes!";
-
+    #[allow(dead_code)]
+    pub const POLICYJSON: [u8; 9] = *b"{\"a\":\"b\"}";
```
fitzthum, line 81: I would probably just use strings for these and then convert to bytes if needed.

fitzthum (commented):
A few comments

src/kbc_modules/offline_fs_kbc/README.md
```diff
@@ -18,6 +18,21 @@ Keys must be provided in the guest file system at `/etc/aa-offline_fs_kbc-keys.j
 
 with the 32-byte keys base64-encoded.
 
+Resources must be provide in the guest file system at `/etc/aa-offline_fs_kbc-resources.json` like:
+```
+{
+  "Policy": "ewogICAgImRlZmF1bHQiOiBbCiAgICAgICAgewogICAgICAgICAgICAidHlwZSI6ICJpbnNlY3VyZUFjY2VwdEFueXRoaW5nIgogICAgICAgIH0KICAgIF0sCiAgICAidHJhbnNwb3J0cyI6IHsKICAgICAgICAiZG9ja2VyIjogewogICAgICAgICAgICAicXVheS5pby9rYXRhLWNvbnRhaW5lcnMiOiBbCiAgICAgICAgICAgICAgICB7CiAgICAgICAgICAgICAgICAgICAgInR5cGUiOiAic2lnbmVkQnkiLAogICAgICAgICAgICAgICAgICAgICJzY2hlbWUiOiAic2ltcGxlIiwKICAgICAgICAgICAgICAgICAgICAia2V5VHlwZSI6ICJHUEdLZXlzIiwKICAgICAgICAgICAgICAgICAgICAia2V5UGF0aCI6ICIvcnVuL2ltYWdlLXNlY3VyaXR5L3NpbXBsZV9zaWduaW5nL3B1YmtleS5ncGciCiAgICAgICAgICAgICAgICB9CiAgICAgICAgICAgIF0KICAgICAgICB9CiAgICB9Cn0=",
```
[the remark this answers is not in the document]
liudalibj: Agree, will remove the full values from the README file, use placeholder here

src/kbc_modules/offline_fs_kbc/README.md
```diff
@@ -18,6 +18,21 @@ Keys must be provided in the guest file system at `/etc/aa-offline_fs_kbc-keys.j
 
 with the 32-byte keys base64-encoded.
 
+Resources must be provide in the guest file system at `/etc/aa-offline_fs_kbc-resources.json` like:
```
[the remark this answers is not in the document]
liudalibj: put everything together to one file here can let us manage the resource contents easily, currently we only support one GPG public key, in the future we will support using multiple GPG public keys when verifying signatures, more details at here https://github.com/confidential-containers/image-rs/issues/50

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -50,9 +51,23 @@ pub fn load_keys(keyfile_name: &str) -> Result<Keys> {
         .collect()
 }
 
+pub fn load_resources(resources_file_name: &str) -> Result<Resources> {
+    let resources_json = fs::read_to_string(resources_file_name)?;
+    // Redact parsing errors to avoid side-channels
```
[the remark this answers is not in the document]
liudalibj: removed this comment here.

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -96,4 +141,25 @@ pub mod tests {
 
         fs::remove_file(keyfile_name).unwrap();
     }
+    #[test]
+    fn test_load_resources() {
+        let resources_file_path = create_resources_file("aa-offline_fs_kbc-test_load_resources");
```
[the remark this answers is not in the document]
liudalibj: append `.json` and use a `const` here

src/kbc_modules/offline_fs_kbc/common.rs
liudalibj: updated to
```
    use crate::kbc_modules::ResourceName::Policy;
    use crate::kbc_modules::ResourceName::SigstoreConfig;
    use crate::kbc_modules::ResourceName::GPGPublicKey;
```

src/kbc_modules/offline_fs_kbc/common.rs
liudalibj: try only import `base64` here, meet follow issue
```
error[E0425]: cannot find function `encode` in this scope
   --> src/kbc_modules/offline_fs_kbc/common.rs:102:17
    |
102 |                 encode(KEY),
    |                 ^^^^^^
    |
   ::: /Users/liudali/.cargo/registry/src/github.com-1ecc6299db9ec823/base64-0.13.0/src/decode.rs:88:1
    |
88  | pub fn decode<T: AsRef<[u8]>>(input: T) -> Result<Vec<u8>, DecodeError> {
    | ----------------------------------------------------------------------- similarly named function `decode` defined here
    |
help: a function with a similar name exists
    |
102 |                 decode(KEY),
    |                 ~~~~~~
help: consider importing one of these items
    |
68  |     use base64::encode;
    |
68  |     use crate::kbc_modules::offline_fs_kbc::common::tests::base64::encode;
    |

```
So keep the code no change
```
pub use base64::encode;
```

src/kbc_modules/offline_fs_kbc/common.rs
liudalibj: updated, remove the `format!` macro and use `json!` macro

src/kbc_modules/offline_fs_kbc/common.rs
liudalibj: updated the codes looks like:
```
    pub fn create_resources_file(resources_file_path: &Path) {
        let resources_file_content = serde_json::json!({
            Policy.to_string(): base64::encode(POLICYJSON.as_bytes()),
            SigstoreConfig.to_string(): base64::encode(SIGSTORECONFIG.as_bytes()),
            GPGPublicKey.to_string(): base64::encode(PUBKEY.as_bytes()),
         });


        fs::write(
            resources_file_path,
            serde_json::to_string(&resources_file_content).unwrap(),
        )
        .unwrap();
    }
```

src/kbc_modules/offline_fs_kbc/common.rs
liudalibj: this function is for test only, so maybe let's keep it in test module?

liudalibj:
@fitzthum Thanks for your review comment, I addressed them would you like help to check this PR again, Thanks?

src/kbc_modules/offline_fs_kbc/common.rs
```diff
@@ -50,10 +51,25 @@ pub fn load_keys(keyfile_name: &str) -> Result<Keys> {
...
+        .collect()
+}
+
 pub mod tests {
     pub use super::*;
-
-    pub use base64::encode;
+    use crate::kbc_modules::ResourceName::GPGPublicKey;
+    use crate::kbc_modules::ResourceName::Policy;
+    use crate::kbc_modules::ResourceName::SigstoreConfig;
```
fitzthum, line 71: I was thinking something slightly different with this import. 

How about `use crate::kbc_modules::ResourceName`

and then when you use Policy below you would do `ResourceName::Policy`

Currently we have something like `Policy.to_string()`, which I think is a bit ambiguous, especially given all the different meanings of Policy across the project.

fitzthum (approved):
LGTM. 

I was thinking something slightly different with one of the changes. See comment.

btw have you tested this with `image-rs`?

fitzthum:
Also, @liudalibj  can you squash your commits into one?

src/kbc_modules/offline_fs_kbc/common.rs
liudalibj: updated

jialez0 (approved):

PR #79, opened 2022-08-28
set verifier_type of rats_tls_conf_t to nullverifier

xhuan28:
When AA works in eaa_kbc mode, it is an attester but not a verifier.  While the current implementation uses the default config, that is the highest priority settings "sgx_ecdsa_qve". The verifier_type of rats_tls_conf_t should be set to nullverifier explicitly, otherwise the remote attestation with verdictd fails.


jialez0 (approved):

jialez0:
@xhuan28 Could you please make the commit log clearer and more concise? They look a little messy now :)

fitzthum (approved):
LGTM, but I agree that the commits could be squashed down. 

Does this mean that with EAA verdictd won't actually verify the evidence?

jialez0:
@fitzthum This parameter does not mean that verdictd will no longer verify the evidence, but specifies that AA will not verify the dummy "evidence" sent by verdictd. It is because the working mode of the component rats-tls that AA depends on. Rats-TLS will enable two-way authentication by default. In our scenario, only verdictd is required to verify AA, and AA is not required to verify verdictd.

@xhuan28 n my test, AA does not need to explicitly specify the nullverifier parameter. You only need to start the verdictd in the following way to establish the link normally:
```
verdictd  --listen 0.0.0.0:30000 --mutual
```

jialez0:
@xhuan28 This problem can be solved through rational configuration of rats-tls, so this PR can be closed.

jialez0:
@xhuan28 This problem can be solved through rational configuration of rats-tls, so this PR can be closed.

PR #82, opened 2022-09-01
Add get_resource support to online_sev_kbc

fitzthum:
Requests resources from KBS by resource name/type.

This is a bit tricky to test end-to-end given that `image-rs` support in Kata isn't upstream yet. I think we will stick with the offline KBS for the first release.

This just uses the name of the resource to make the request, but it looks like we're really treating the name of the resource more like the type of the resource. I suspect we might want to add support for an additional resource id in the future. 

@dubek 

Signed-off-by: Tobin Feldman-Fitzthum <tobin@ibm.com>

### DIFF
--- a/src/kbc_modules/online_sev_kbc/mod.rs
+++ b/src/kbc_modules/online_sev_kbc/mod.rs
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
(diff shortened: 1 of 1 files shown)


jodh-intel (approved):
Thanks @fitzthum.

lgtm

dubek (approved):
LGTM.

Small possible style comment below.

sameo (approved):

PR #90, opened 2022-10-04
Extract common SEV code

fitzthum:
This pulls out some common code that was duplicated between `offline_sev_kbc` and `online_sev_kbc`.

In the past we have put common code into separate files in the KBC modules. That never really worked very well so I created a new module for shared components that should be a bit more flexible.

Signed-off-by: Tobin Feldman-Fitzthum <tobin@ibm.com>


dubek (approved):
Looks good to me -- just extracts common code, no functional change intended.

PR #97, opened 2022-10-25
EAA KBC: Fix compile bug of rats-tls callback on TDX.

jialez0:
EAA KBC does not use the callback function of rats-tls, so it can be set to None directly. (If do not do this, errors will occur when compiling on the TDX platform.)


fitzthum (approved):
LGTM

PR #104, opened 2022-11-14
Update cc-kbc to follow KBS changes 

### LINKED ISSUE #102: Update cc-kbc to match KBS
There have been some updates to the KBS protocol, such as removing the token from the request message, renaming evidence to attestation, etc. Is there any plan to sync up the `cc-kbc` with these new changes? We could potentially use the `kbs-types` crate, although that might not fit with some of the existing crypto stuff.

I am hoping to start pushing more on SNP enablement, which will require the`cc-kbc` to actually make requests to the KBS. If nobody else is interested in updating the `cc-kbc` I will start working on it next week.

### LINKED ISSUE #224: Update cc-kbc to match KBS
There have been some updates to the KBS protocol, such as removing the token from the request message, renaming evidence to attestation, etc. Is there any plan to sync up the `cc-kbc` with these new changes? We could potentially use the `kbs-types` crate, although that might not fit with some of the existing crypto stuff.

I am hoping to start pushing more on SNP enablement, which will require the`cc-kbc` to actually make requests to the KBS. If nobody else is interested in updating the `cc-kbc` I will start working on it next week.

fitzthum:
Resolves confidential-containers/guest-components#224 

@dubek @jialez0 

There are still no tests for this. Either we try to write some now or we do it later when we figure out what we are doing for a KBS.


src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -70,59 +69,53 @@ impl TeeKey {
...
-                    .decrypt(nonce, cipher_text.as_ref())
-                    .map_err(|e| anyhow!("AES_GCM_256_ALGORITHM: {:?}", e))?
-            }
-            _ => {
-                return Err(anyhow!("Unsupported algorithm: {}", self.algorithm.clone()));
-            }
-        };
-
-        Ok(plain_text)
+pub fn decrypt_response(response: &Response, tee_key: TeeKey) -> Result<Vec<u8>> {
```
dubek, line 80: consider making this function a method on Response:

```rust
impl Response {
    fn decrypt(self, tee_key: TeeKey) -> .... {
         ...
    }
}
```

(but maybe that'll make you move it away from the `crypto` mod. not sure)

Cargo.toml
```diff
@@ -29,6 +29,7 @@ tokio = { version = "1.0", features = ["macros", "rt-multi-thread"], optional =
 uuid = { version = "1.1.2", features = ["serde", "v4"], optional = true }
 bincode = { version = "1.3.3", optional = true }
 futures-util = "0.3.21"
+base64-url = "1.4.13"
```
dubek, line 32: if we don't want to introduce a new crate (we already use `base64` above), we can use the existing crate with:

```rust
let b64_url = base64::encode_config(b"hello internet~", base64::URL_SAFE);
```

dubek (approved):
Looks good to me.

Small questions about code organization below:

src/kbc_modules/cc_kbc/crypto/mod.rs
fitzthum: Yeah this was just to keep things together in the crypto file. Might be able to impl from there if I make a bunch of stuff public.

Cargo.toml
dubek: The `base64` crate is much more established (just looking at github stars, contributors, and crates.io downloads count). So my preference is to switch the calls to `base64::encode_config()`, and remove the new dependency.

BTW we should use config `base64::URL_SAFE_NO_PAD` and not like I wrote above.

src/kbc_modules/cc_kbc/crypto/mod.rs
dubek: Just leave it in `crypto.rs`.

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -11,6 +12,7 @@ use rsa::{PaddingScheme, RsaPrivateKey, RsaPublicKey};
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha384};
 
+const RSA_KEY_TYPE: &str = "RSA";
 const RSA_ALGORITHM: &str = "rsa-pkcs1v15";
```
jialez0, line 16: Here I think we should use`RSA1_5` follow Section 4.1 of [rfc7518](https://www.rfc-editor.org/rfc/inline-errata/rfc7518.html), and so as bellow (`AES_GCM_256_ALGORITHM`)

jialez0 (approved):
Thanks @fitzthum ! This is a nice work. In addition, KBS now defines how to use the https API to communicate with it, so we can implement the module to send https requests to KBS in our `cc-kbc`, but this PR is a good start.

src/kbc_modules/cc_kbc/crypto/mod.rs
jialez0: We may should also change `const AES_GCM_256_ALGORITHM: &str = "aes-gcm-256";` to `const AES_GCM_256_ALGORITHM: &str = "A256GCM"` to follow Section 5.3 of [rfc7518](https://www.rfc-editor.org/rfc/inline-errata/rfc7518.html).

fitzthum:
Yeah next step will be to add in the API calls according to the KBS spec.

PR #105, opened 2022-11-15
Add simple SNP attester

fitzthum:
Just a draft for a simple SNP attester. This doesn't use the kernel hashes stuff, but it's a good starting point. 

This PR depends on https://github.com/virtee/sev/pull/45 which is still in flux so things will change. 

@dubek

### DIFF
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -39,6 +39,7 @@ cc_kbc = ["attestation_agent/cc_kbc", "attestation_agent/all-attesters"]
 cc_kbc_tdx = ["attestation_agent/cc_kbc", "attestation_agent/tdx-attester"]
 cc_kbc_occlum = ["attestation_agent/cc_kbc","attestation_agent/occlum-attester"]
 cc_kbc_az_snp_vtpm = ["attestation_agent/cc_kbc", "attestation_agent/az-snp-vtpm-attester"]
+cc_kbc_snp = ["attestation_agent/cc_kbc", "attestation_agent/snp-attester"]
 
 eaa_kbc = ["attestation_agent/eaa_kbc"]
 offline_fs_kbc = ["attestation_agent/offline_fs_kbc"]
--- a/attester/Cargo.toml
+++ b/attester/Cargo.toml
@@ -13,13 +13,15 @@ log.workspace = true
 occlum_dcap = { git = "https://github.com/occlum/occlum", rev = "dbe404f", optional = true }
 serde.workspace = true
 serde_json.workspace = true
+sev = { git = "https://github.com/virtee/sev", rev = "3dca05d2c93388cb00534ad18f5928fd812e99cc", optional = true }
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", rev = "cc582e8be0c9010295c66fb58c59f74744017600", optional = true }
 
 [features]
 default = ["all-attesters"]
-all-attesters = ["tdx-attester", "occlum-attester", "az-snp-vtpm-attester"]
+all-attesters = ["tdx-attester", "occlum-attester", "az-snp-vtpm-attester", "snp-attester"]
 
 tdx-attester = ["tdx-attest-rs"]
 occlum-attester = ["occlum_dcap"]
 az-snp-vtpm-attester = ["az-snp-vtpm"]
+snp-attester = ["sev"]
--- a/attester/src/lib.rs
+++ b/attester/src/lib.rs
@@ -12,17 +12,21 @@ pub mod sample;
 
 #[cfg(feature = "az-snp-vtpm-attester")]
 pub mod az_snp_vtpm;
+
 #[cfg(feature = "tdx-attester")]
 pub mod tdx;
 
 #[cfg(feature = "occlum-attester")]
 pub mod sgx_occlum;
 
+#[cfg(feature = "snp-attester")]
+pub mod snp;
+
 /// The supported TEE types:
 /// - Tdx: TDX TEE.
 /// - SgxOcclum: SGX TEE with Occlum Libos.
 /// - AzSnpVtpm: SEV-SNP TEE for Azure CVMs.
-/// - Sevsnp: SEV-SNP TEE.
+/// - Snp: SEV-SNP TEE.
 /// - Sample: A dummy TEE that used to test/demo the KBC functionalities.
 #[derive(Debug, EnumString, Display)]
 #[strum(ascii_case_insensitive, serialize_all = "lowercase")]
@@ -30,8 +34,8 @@ pub enum Tee {
     Tdx,
     #[strum(serialize = "sgx")]
     SgxOcclum,
-    Sevsnp,
     AzSnpVtpm,
+    Snp,
     Sample,
     Unknown,
 }
@@ -46,6 +50,8 @@ impl Tee {
             Tee::SgxOcclum => Ok(Box::<sgx_occlum::SgxOcclumAttester>::default()),
             #[cfg(feature = "az-snp-vtpm-attester")]
             Tee::AzSnpVtpm => Ok(Box::<az_snp_vtpm::AzSnpVtpmAttester>::default()),
+            #[cfg(feature = "snp-attester")]
+            Tee::Snp => Ok(Box::<snp::SnpAttester>::default()),
             _ => bail!("TEE is not supported!"),
         }
     }
@@ -76,5 +82,9 @@ pub fn detect_tee_type() -> Tee {
         return Tee::AzSnpVtpm;
     }
 
+    #[cfg(feature = "snp-attester")]
+    if snp::detect_platform() {
+        return Tee::Snp;
+    }
     Tee::Unknown
 }
--- a/attester/src/snp/mod.rs
+++ b/attester/src/snp/mod.rs
@@ -1,0 +1,51 @@
+// Copyright (c) 2022 IBM
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use super::Attester;
+use anyhow::*;
+use serde::{Deserialize, Serialize};
+use sev::firmware::guest::types::{AttestationReport, SnpReportReq};
+use sev::firmware::guest::Firmware;
+use sev::firmware::host::types::CertTableEntry;
+use std::path::Path;
+
+pub fn detect_platform() -> bool {
+    Path::new("/sys/devices/platform/sev-guest").exists()
+}
+
+#[derive(Serialize, Deserialize)]
+struct SnpEvidence {
+    attestation_report: AttestationReport,
+    cert_chain: Vec<CertTableEntry>,
+}
+
+#[derive(Debug, Default)]
+pub struct SnpAttester {}
+
+impl Attester for SnpAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data)?;
+
+        if report_data_bin.len() != 48 {
+            bail!("Malformed SNP Evidence");
+        }
+
+        report_data_bin.extend([0; 16]);
+
+        let mut firmware = Firmware::open()?;
+        let report_request = SnpReportReq::new(Some(report_data_bin.as_slice().try_into()?), 0);
+
+        let (report, certs) = firmware
+            .snp_get_ext_report(None, report_request)
+            .context("Failed to get attestation report")?;
+
+        let evidence = SnpEvidence {
+            attestation_report: report,
+            cert_chain: certs,
+        };
+
+        serde_json::to_string(&evidence).context("Serialize SNP evidence failed")
+    }
+}
--- a/.github/workflows/cc_kbc.yml
+++ b/.github/workflows/cc_kbc.yml
@@ -30,6 +30,7 @@ jobs:
           - cc_kbc_tdx
           - cc_kbc_occlum
           - cc_kbc_az_snp_vtpm
+          - cc_kbc_snp
     steps:
       - name: Code checkout
         uses: actions/checkout@v2
--- a/kbc/Cargo.toml
+++ b/kbc/Cargo.toml
@@ -41,6 +41,7 @@ all-attesters = ["kbs_protocol/all-attesters"]
 tdx-attester = ["kbs_protocol/tdx-attester"]
 occlum-attester = ["kbs_protocol/occlum-attester"]
 az-snp-vtpm-attester= ["kbs_protocol/az-snp-vtpm-attester"]
+snp-attester = ["kbs_protocol/snp-attester"]
 
 sample_kbc = []
 eaa_kbc = ["foreign-types"]
--- a/kbs_protocol/Cargo.toml
+++ b/kbs_protocol/Cargo.toml
@@ -25,6 +25,7 @@ all-attesters = ["attester/all-attesters"]
 tdx-attester = ["attester/tdx-attester"]
 occlum-attester = ["attester/occlum-attester"]
 az-snp-vtpm-attester = ["attester/az-snp-vtpm-attester"]
+snp-attester = ["attester/snp-attester"]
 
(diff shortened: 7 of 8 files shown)


fitzthum:
There has been some discussion about the best form to store the Attestation Report and Certs in as part of the evidence. I am leaning towards just putting them in as Structs and letting the json serialization do its thing. Currently we turn them into binary, and then base64, and then put them into JSON. This is redundant but maybe it's a bit more portable.

src/kbc_modules/cc_kbc/attester/snp/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+        let (report, certs) = firmware
+            .snp_get_ext_report(None, extended_report_request)
+            .map_err(|e| anyhow!("failed to get attestation report: {:?}", e))?;
+
+        let evidence = SnpEvidence {
+            attestation_report: base64::encode(bincode::serialize(&report)?),
+            cert_chain: base64::encode(bincode::serialize(&certs)?),
+        };
+
+        serde_json::to_string(&evidence).map_err(|_| anyhow!("Serialize sample evidence failed"))
```
dubek, line 50: s/sample/SNP/

Cargo.toml
```diff
@@ -27,8 +27,10 @@ tonic = { version = "0.8.0", optional = true }
 prost = { version = "0.11.0", optional = true }
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread"], optional = true }
 uuid = { version = "1.1.2", features = ["serde", "v4"], optional = true }
-bincode = { version = "1.3.3", optional = true }
+bincode = { version = "1.3.3" }
 futures-util = "0.3.21"
+sev = { version = "1.0.0", optional = true, git = "https://github.com/larrydewey/virtee-sev.git", branch = "43-Clean-Extended-Report" }
+raw-cpuid = "10.6.0"
```
dubek, line 33: should `raw-cpuid` be `optional = true` as well, and listed below as a dependency for specific feature `cc_kbc` ?

src/kbc_modules/cc_kbc/attester/snp/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Debug, Default)]
+pub struct SnpAttester {}
+
+impl Attester for SnpAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data)?;
+        assert_eq!(report_data_bin.len(), 48);
+        report_data_bin.extend([0; 16]);
+
+        let mut firmware = Firmware::open()?;
```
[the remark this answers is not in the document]
fitzthum: Yeah I think we can do this with Kata's kernel module loading feature.

fitzthum:
https://github.com/virtee/sev/pull/45 has been merged now btw

Cargo.toml
```diff
@@ -40,8 +42,8 @@ shadow-rs = "0.16.1"
 [features]
 default = ["sample_kbc"]
 sample_kbc = []
-cc_kbc = []
+cc_kbc = ["sev", "raw-cpuid"]
```
Xynnn007, line 45: I'm not sure whether this could cause the same issue https://github.com/confidential-containers/attestation-agent/pull/116 tries to fix in tdx. For example, when a user wants to build cc-kbc on a non-sev machine, the default feature `sev` here for `cc_kbc` might try to find sev-architecture-specific dependencies (especially c-rust calls), which might let the build fail.

fitzthum: You should be able to build `sev` crate on any KVM host. Btw if we need TDX to build the TDX attester, how are we going to build it on the CI or when packaging the initrd?

Xynnn007: Yes, you're right. The key point I want here is that some dependencies should also be installed to help the compile process, although the machine used for CI doesn't support that kind of TEE. This seems weird at the first glance, but it is reasonable for we can use the compiled binary on different kinds of attester machines out-of-the-box-ly.

Cargo.toml
```diff
@@ -27,8 +27,10 @@ tonic = { version = "0.8.0", optional = true }
 prost = { version = "0.11.0", optional = true }
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread"], optional = true }
 uuid = { version = "1.1.2", features = ["serde", "v4"], optional = true }
-bincode = { version = "1.3.3", optional = true }
+bincode = { version = "1.3.3" }
 futures-util = "0.3.21"
+sev = { version = "1.0.1", optional = true }
```
larrydewey, line 32: It would be a really good idea to bump this to:

```toml
sev = { version = "1.1.0", optional = true }
```

As the extended report patches and the asked changes for public visibility were applied to that release.

src/kbc_modules/cc_kbc/attester/snp/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use super::Attester;
+use anyhow::*;
+use serde::{Deserialize, Serialize};
+use sev::firmware::guest::types::{AttestationReport, SnpReportReq};
+use sev::firmware::guest::Firmware;
+use sev::firmware::host::types::CertTable;
```
larrydewey, line 11: With the suggested changes you provided for `virtee/sev` in version `1.1.0`, this will be replaced with a `Vec<CertTableEntry>`

src/kbc_modules/cc_kbc/attester/snp/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+use std::path::Path;
+
+pub fn detect_platform() -> bool {
+    Path::new("/sys/devices/platform/sev-guest").exists()
+}
+
+#[derive(Serialize, Deserialize)]
+struct SnpEvidence {
+    attestation_report: AttestationReport,
+    cert_chain: CertTable,
```
larrydewey, line 21: Same modification here from the change above (`Vec<CertTableEntry>`).

src/kbc_modules/cc_kbc/attester/snp/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[derive(Debug, Default)]
+pub struct SnpAttester {}
+
+impl Attester for SnpAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data)?;
+        assert_eq!(report_data_bin.len(), 48);
```
larrydewey, line 30: It might be beneficial to add a message for if/when this fails.

larrydewey (changes requested):
Just a couple of additional changes due to the changes in the latest revision of `virtee/sev`

src/kbc_modules/cc_kbc/attester/snp/mod.rs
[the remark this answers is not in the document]
jepio: We could also require that the `sev-guest` module is built-in.

Cargo.toml
```diff
@@ -27,8 +27,10 @@ tonic = { version = "0.8.0", optional = true }
 prost = { version = "0.11.0", optional = true }
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread"], optional = true }
 uuid = { version = "1.1.2", features = ["serde", "v4"], optional = true }
```
jepio, line 29: We need to bump uuid because virtee/sev requires `^1.2` (https://github.com/virtee/sev/blob/v1.1.0/Cargo.toml#L40).

fitzthum: Shouldn't virtee/sev pull in its own dependencies? Or is there some incompatibility between the versions?

fitzthum: It builds for me. That's kind of odd. No problem to update it.

fitzthum:
btw this has been rebased and my branch now includes all the TDX stuff, meaning that we will need something like https://github.com/kata-containers/kata-containers/pull/6373 to build it

src/kbc_modules/cc_kbc/attester/mod.rs
```diff
@@ -20,7 +23,7 @@ pub mod tdx;
 pub enum Tee {
     Tdx,
     Sgx,
-    Sevsnp,
+    Snp,
```
dubek, line 26: Good; don't forget to modify the comment as well (line 19).

Cargo.toml
```diff
@@ -44,6 +47,7 @@ default = ["sample_kbc", "rust-crypto"]
 cc_kbc = ["rand", "rsa", "sha2", "reqwest"]
 all-attesters = ["tdx-attester"]
 tdx-attester = ["tdx-attest-rs"]
+snp-attester = ["sev", "raw-cpuid"]
```
dubek, line 50: add `"snp-attester"` to `all-attesters`.

src/kbc_modules/cc_kbc/attester/snp/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data)?;
+        assert_eq!(report_data_bin.len(), 48, "Malformed Evidence");
+        report_data_bin.extend([0; 16]);
+
+        let mut firmware = Firmware::open()?;
+        let mut report_request = SnpReportReq::new(Some(report_data_bin.as_slice().try_into()?), 0);
+
+        let (report, certs) = firmware
+            .snp_get_ext_report(None, &mut report_request)
```
dubek, line 37: I think there's no need to pass a `&mut` and also `report_request` doesn't need to be `mut`.
(by looking at latest code in virtee/sev main branch; maybe we use an outdated version?)

fitzthum: This change isn't in any release yet, but I"ll keep an eye on it.

jepio, line 37: I just realized that some hypervisors that are used to run SNP guests might not support `snp_get_ext_report()`, so maybe a fallback to a plain `snp_get_report()` would be nice?

fitzthum: Yeah that's probably a good idea. Should the verifier then also have a fallback where it contacts the KDS?

jepio: that sounds like it might be a good idea as well, yeah

fitzthum:
The time has finally come for this to no longer be a draft. We have tested internally with the corresponding verifier PR and it works. We don't really have CI for SNP (especially with attestation) so tbd exactly when this will be supported, but this code is more or less ready.

There are a few outstanding comments that I will look into.

attester/Cargo.toml
```diff
@@ -15,11 +15,14 @@ serde.workspace = true
 serde_json.workspace = true
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", rev = "cc582e8be0c9010295c66fb58c59f74744017600", optional = true }
+sev = { git = "https://github.com/virtee/sev", rev = "3dca05d2c93388cb00534ad18f5928fd812e99cc", optional = true }
+raw-cpuid = { version = "10.6.0", optional = true }
```
dubek, line 19: where is `raw-cpuid` used? if not needed, remove here and in line 28.

attester/src/lib.rs
```diff
@@ -12,26 +12,31 @@ pub mod sample;
...
 /// - Sample: A dummy TEE that used to test/demo the KBC functionalities.
 #[derive(Debug, EnumString, Display)]
 #[strum(ascii_case_insensitive, serialize_all = "lowercase")]
 pub enum Tee {
     Tdx,
     #[strum(serialize = "sgx")]
     SgxOcclum,
-    Sevsnp,
     AzSnpVtpm,
+    Sgx,
```
dubek, line 38: there's an `SgxOcclum` in line 36; do we need this as well?

dubek (commented):
Almost perfect, few small comments below.

attester/src/lib.rs
Xynnn007: IMO, we do not need this.

attester/Cargo.toml
fitzthum: Ah yeah I think that is left over from an earlier approach for checking the platform type.

src/kbc_modules/cc_kbc/attester/snp/mod.rs
fitzthum: Reasonable to add these fallbacks in a future PR?

fitzthum:
Ok, @dubek @jepio @Xynnn007 any more comments?

attester/Cargo.toml
```diff
@@ -15,11 +15,13 @@ serde.workspace = true
 serde_json.workspace = true
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", rev = "cc582e8be0c9010295c66fb58c59f74744017600", optional = true }
+sev = { git = "https://github.com/virtee/sev", rev = "3dca05d2c93388cb00534ad18f5928fd812e99cc", optional = true }
```
Xynnn007, line 18: Please move this `sev` following the lexicographical order

app/Cargo.toml
```diff
@@ -39,7 +39,7 @@ cc_kbc = ["attestation_agent/cc_kbc", "attestation_agent/all-attesters"]
 cc_kbc_tdx = ["attestation_agent/cc_kbc", "attestation_agent/tdx-attester"]
 cc_kbc_occlum = ["attestation_agent/cc_kbc","attestation_agent/occlum-attester"]
 cc_kbc_az_snp_vtpm = ["attestation_agent/cc_kbc", "attestation_agent/az-snp-vtpm-attester"]
-
+cc_kbc_snp = ["attestation_agent/cc_kbc", "attestation_agent/snp-attester"]
```
Xynnn007, line 42: nit: could we add an empty line between `cc_kbc_*` and `eaa_kbc`?

attester/src/snp/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    cert_chain: Vec<CertTableEntry>,
+}
+
+#[derive(Debug, Default)]
+pub struct SnpAttester {}
+
+impl Attester for SnpAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data)?;
+        assert_eq!(report_data_bin.len(), 48, "Malformed Evidence");
```
Xynnn007, line 30: If we use `assert_eq!` here, the error will cause the program to `panic`. How about a raised error? s.t.
```rust
if report_data_bin.len() != 48 {
    bail!("Malformed Evidence of SNP");
}
```

This prevents panic inside

attester/src/snp/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+        let (report, certs) = firmware
+            .snp_get_ext_report(None, report_request)
+            .map_err(|e| anyhow!("failed to get attestation report: {:?}", e))?;
+
+        let evidence = SnpEvidence {
+            attestation_report: report,
+            cert_chain: certs,
+        };
+
+        serde_json::to_string(&evidence).map_err(|_| anyhow!("Serialize SNP evidence failed"))
```
Xynnn007, line 45: just nit:
`anyhow` provides a simple way to add context information for an error:
```rust
serde_json::to_string(&evidence).context("Serialize SNP evidence failed")
```

not sure whether it is proper here

Xynnn007 (approved):
Thanks, only some nits

dubek (approved):
Looks good!

src/kbc_modules/cc_kbc/attester/snp/mod.rs
jepio: Sure

jepio (approved):

fitzthum:
@jialez0 or @sameo PTAL

jialez0 (approved):
LGTM.

PR #106, opened 2022-11-16
Introduce Client Id resource 

fitzthum:
The Client Id resource type allows a guest to request an Id that is provided by the KBS. This can allow a client to more easily associate a particular KBC/workload with a KBS. This value has a somewhat limited use because it cannot be shared publicly. Nonetheless it seems handy and is trivial to implement for SEV.

@dubek

### DIFF
--- a/src/kbc_modules/online_sev_kbc/mod.rs
+++ b/src/kbc_modules/online_sev_kbc/mod.rs
@@ -4,17 +4,18 @@
 //
 
 use crate::common::sev::*;
-use crate::kbc_modules::{KbcCheckInfo, KbcInterface, ResourceDescription};
+use crate::kbc_modules::{KbcCheckInfo, KbcInterface, ResourceDescription, ResourceName};
 
 use aes_gcm::aead::{Aead, KeyInit};
 use aes_gcm::{Aes256Gcm, Key, Nonce};
-use anyhow::{anyhow, Result};
+use anyhow::*;
 use async_trait::async_trait;
 use base64::decode;
 use openssl::symm::{decrypt, Cipher};
 use serde::Deserialize;
 use std::collections::HashMap;
 use std::fs;
+use std::str::FromStr;
 use tonic::codegen::http::Uri;
 use uuid::Uuid;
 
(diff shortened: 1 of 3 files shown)


src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -78,7 +79,16 @@ impl KbcInterface for OnlineSevKbc {
 
     async fn get_resource(&mut self, description: String) -> Result<Vec<u8>> {
         let desc = serde_json::from_str::<ResourceDescription>(description.as_str())?;
-        self.get_resource_from_kbs(desc.name).await
+        match ResourceName::from_str(desc.name.as_str())? {
```
dubek, line 82: what happens if `desc.name` is just garbage? (none of the expected ResourceName enum values)

previously, this was passed to `get_resource_from_kbs()` which would then just send a query to the KBS with `secret_type="resource"` and `id="garbage"`.

now: `ResourceName::from_str("garbage")` might return Err and the KBS will not be queried.

this is a change in behaviour?

Another possible change:



maybe the definition of `ResourceDescription` should have field `name` with type `ResourceName` instead of `String`. this will have the effect but clearer declarations.

src/kbc_modules/online_sev_kbc/mod.rs
```diff
@@ -78,7 +79,16 @@ impl KbcInterface for OnlineSevKbc {
...
+            ResourceName::ClientId => {
+                let connection = self
+                    .connection
+                    .as_ref()
+                    .map_err(|e| anyhow!("Failed to get injected connection. {}", e))?;
+                Ok(connection.client_id.hyphenated().to_string().into_bytes())
```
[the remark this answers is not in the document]
fitzthum: Sadly the `hyphenated` struct can't be turned into bytes directly.

src/kbc_modules/online_sev_kbc/mod.rs
fitzthum: Tweaked this slightly to preserve the current behavior although having a filter at the KBC might actually be a good idea.

dubek (approved):
LGTM

jialez0 (approved):
LGTM.

PR #110, opened 2022-11-28
CC-KBC: Add TDX Attester

### LINKED ISSUE #220: TDX Support in CC-KBC.
The generic CC-KBC needs to support TDX platform, which means that we need to add `tdx` attester to support the following functions:

1. Check whether the current platform is based on TDX hardware.
2. Get TDX quote.
3. Read the measurement event log which contains measurement of kernel, kernel params.

jialez0:
Resolve: confidential-containers/guest-components#220

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -57,4 +57,4 @@ jobs:
         with:
           command: clippy
           # We are getting error in generated code due to derive_partial_eq_without_eq check, so ignore it for now
-          args: --all-targets --all-features -- -D warnings -A clippy::derive_partial_eq_without_eq
+          args: --all-targets -- -D warnings -A clippy::derive_partial_eq_without_eq
--- a/src/kbc_modules/cc_kbc/attester/tdx/mod.rs
+++ b/src/kbc_modules/cc_kbc/attester/tdx/mod.rs
@@ -1,0 +1,83 @@
+// Copyright (c) 2022 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use super::Attester;
+use anyhow::*;
+use serde::{Deserialize, Serialize};
+use std::path::Path;
+use tdx_attest_rs;
+
+const CCEL_PATH: &str = "/sys/firmware/acpi/tables/CCEL";
+
+pub fn detect_platform() -> bool {
+    Path::new("/dev/tdx-attest").exists()
+}
+
+#[derive(Serialize, Deserialize)]
+struct TdxEvidence {
+    // Base64 encoded CC Eventlog ACPI table
+    // refer to https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#cc-event-log-acpi-table.
+    cc_eventlog: Option<String>,
+    // Base64 encoded TD quote.
+    quote: String,
+}
+
+#[derive(Debug, Default)]
+pub struct TdxAttester {}
+
+impl Attester for TdxAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data)?;
+        if report_data_bin.len() != 48 {
+            return Err(anyhow!(
+                "TDX Attester: Report data should be SHA384 base64 String"
+            ));
+        }
+        report_data_bin.extend([0; 16]);
+
+        let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
+            d: report_data_bin.as_slice().try_into()?,
+        };
+
+        let quote = match tdx_attest_rs::tdx_att_get_quote(Some(&tdx_report_data), None, None, 0) {
+            (tdx_attest_rs::tdx_attest_error_t::TDX_ATTEST_SUCCESS, Some(q)) => base64::encode(q),
+            (error_code, _) => {
+                return Err(anyhow!(
+                    "TDX Attester: Failed to get TD quote. Error code: {:?}",
+                    error_code
+                ));
+            }
+        };
+
+        let cc_eventlog = match std::fs::read(CCEL_PATH) {
+            Result::Ok(el) => Some(base64::encode(el)),
+            Result::Err(e) => {
+                log::warn!("Read CC Eventlog failed: {:?}", e);
+                None
+            }
+        };
+
+        let evidence = TdxEvidence { cc_eventlog, quote };
+
+        serde_json::to_string(&evidence)
+            .map_err(|e| anyhow!("Serialize TDX evidence failed: {:?}", e))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[ignore]
+    #[test]
+    fn test_tdx_get_evidence() {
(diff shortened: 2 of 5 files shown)


src/kbc_modules/cc_kbc/attester/tdx/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+            ));
+        }
+        report_data_bin.extend([0; 16]);
+
+        let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
+            d: report_data_bin.as_slice().try_into()?,
+        };
+
+        let mut selected_att_key_id = tdx_attest_rs::tdx_uuid_t { d: [0; 16usize] };
+        let (get_quote_result, quote) = tdx_attest_rs::tdx_att_get_quote(
```
dubek, line 45: if it's possible to modify tdx-attest-rs to be more "rusty" and return something like `Result<Vec<u8>, TdxErrorCode>` it would make the code here a bit nicer...

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+                "TDX Attester: Report data should be SHA384 base64 String"
+            ));
+        }
+        report_data_bin.extend([0; 16]);
+
+        let tdx_report_data = tdx_attest_rs::tdx_report_data_t {
+            d: report_data_bin.as_slice().try_into()?,
+        };
+
+        let mut selected_att_key_id = tdx_attest_rs::tdx_uuid_t { d: [0; 16usize] };
```
dubek, line 44: this is given to `tdx_att_get_quote()` but the returned value is not used or returned to the user.
can you please explain its purpose?

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+
+pub fn detect_platform() -> bool {
+    Path::new("/dev/tdx-attest").exists()
+}
+
+#[derive(Serialize, Deserialize)]
+struct TdxEvidence {
+    // Base64 encoded CC Eventlog ACPI table
+    // refer to https://uefi.org/specs/ACPI/6.5/05_ACPI_Software_Programming_Model.html#cc-event-log-acpi-table.
+    cc_eventlog: Option<String>,
```
dubek, line 22: Why is this Option? I see that if there's an error reading the CCEL file the whole `get_evidence()` function will return with error (which sounds like the good approach to me). So it seems that there should always be a value for `cc_eventlog`.

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+use super::Attester;
+use anyhow::*;
+use serde::{Deserialize, Serialize};
+use std::path::Path;
+use tdx_attest_rs;
+
+const CCEL_PATH: &str = "/sys/firmware/acpi/tables/data/CCEL";
+
+pub fn detect_platform() -> bool {
+    Path::new("/dev/tdx-attest").exists()
```
dubek, line 15: will this device auto-load in TDX VMs?
do you need to add it to some kata-gent config so it will be loaded by the agent?
or is it compiled as a built-in (`CONFIG...=y`) in the tdx kernel?

dubek (commented):
Looks good and simple. 
A few questions:

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
jialez0: I agree. But this library is referenced from https://github.com/intel/SGXDataCenterAttestationPrimitives/blob/master/QuoteGeneration/quote_wrapper/tdx-attest-rs, I will feed this suggestion back to Intel developers, but at present, we can only use their current API format...

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
jialez0: Sure, I will add comment here to explain this.

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
jialez0: This device will be automatically included and loaded in all guest kernel that support TDX. This is the core device for TDX attestation(If no `/dev/tdx-attest`, app can not perform TD Quote getting.), so it is appropriate to use it to detect the platform

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            return Err(anyhow!(
+                "TDX Attester: Failed to get TD quote. Error code: {:?}",
+                get_quote_result
+            ));
+        }
+
+        let cc_eventlog = match std::fs::read(CCEL_PATH) {
+            Result::Ok(el) => Some(base64::encode(el)),
+            Result::Err(_) => None,
```
dubek, line 57: should we log (`warn!()`) in this case so we know something is not exactly right?

jialez0: Resolved.

src/kbc_modules/cc_kbc/attester/tdx/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+    // Base64 encoded TD quote.
+    quote: String,
+}
+
+#[derive(Debug, Default)]
+pub struct TdxAttester {}
+
+impl Attester for TdxAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let mut report_data_bin = base64::decode(report_data.clone())?;
```
fitzthum, line 32: Fine for now, but in the future we might want to change `report_data` to be a `[u8; 48]` rather than a String.

fitzthum (approved):
LGTM

dubek (approved):
LGTM, one style suggestion below:

.github/workflows/basic.yml
```diff
@@ -28,6 +28,13 @@ jobs:
...
           sudo apt-get update && sudo apt-get install -y protobuf-compiler
+      
+      - name: Install TDX dependencies
+        run: |
+          sudo curl -L https://download.01.org/intel-sgx/sgx_repo/ubuntu/intel-sgx-deb.key | sudo apt-key add -
+          sudo echo 'deb [arch=amd64] https://download.01.org/intel-sgx/sgx_repo/ubuntu focal main' | sudo tee /etc/apt/sources.list.d/intel-sgx.list
+          sudo apt-get update
+          sudo apt-get install -y libtdx-attest-dev
 
       - name: Build and install with default features
```
dubek, line 39: hmmm -- don't we need a new `.github/workflows/cc_kbc.yml` to build with `make KBC=cc_kbc` ?

I'm not convinced the current CI check on this PR actually compiles the code of the TDX attester.

jialez0: `cc_kbc.yml` is a good idea.

PR #116, opened 2022-12-26
fix: bring tdx attester to a feature

Xynnn007:
Now `cc-kbc` feature includes `tdx-attest-rs` dependency by default. When a user does not install DCAP QGL lib (tdx quote generation lib), the compiler will report error `cannot find tdx_attest.h`. This commit wraps those code who use QGL inside a conditional compile code block, s.t. `tdx`, thus let users choose their own target platform when building the AA binary.


Xynnn007:
The CI error here seems not related to this PR

PR #122, opened 2023-01-10
CC-KBC: Implement KBS resource request

jialez0:


src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -48,10 +72,12 @@ impl Kbc {
 
         Kbc {
             tee: tee_type.to_string(),
+            kbs_uri,
             token: None,
             nonce: String::default(),
             tee_key: TeeKey::new().ok(),
             attester,
+            kbs_client: reqwest::Client::builder().cookie_store(true).build().ok(),
```
dubek, line 80: 1. I suggest setting user-agent (`attestation-agent-cc-kbc/VERSION`) to help with analyzing access logs in the KBS.
2. I suggest setting `.timeout(Duration::from_secs(30))` (the default is no timeout)
3. Since this becomes a very long line, maybe extract this to a `fn build_kbs_client()`.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -17,10 +17,12 @@ use kbs_protocol::message::*;
 
 pub struct Kbc {
     tee: String,
+    kbs_uri: String,
     token: Option<String>,
     nonce: String,
     tee_key: Option<TeeKey>,
     attester: Option<Box<dyn Attester + Send + Sync>>,
+    kbs_client: Option<reqwest::Client>,
```
dubek, line 25: why is this an Option?
can this Kbc do anything without a properly initialized `kbs_client` ?

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +120,60 @@ impl Kbc {
...
+            let attest_response = self.kbs_client()?
+                .post(format!("{}/attest", kbs_uri))
+                .header("Content-Type", "application/json")
+                .body(serde_json::to_string(&attestation_evidence)?)
+                .send()
+                .await?;
+            if attest_response.status() != reqwest::StatusCode::OK {
+                // TODO: Parse Error description payload
+                return Err(anyhow!("KBS attest failed"));
+            }
```
dubek, line 167: I suggest extracting this (/auth and /attest requests) to a separate fn (establish_session ?).

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +120,60 @@ impl Kbc {
...
+    fn kbs_client(&mut self) -> Result<&mut reqwest::Client> {
+        if let Some(kbs_client) = &mut self.kbs_client {
+            Ok(kbs_client)
+        } else {
+            Err(anyhow!("Start kbs http client failed"))
+        }
+    }
+
+    async fn request_kbs_resource(&mut self, resource_url: String) -> Result<Response> {
+        let mut res = self.kbs_client()?.get(&resource_url).send().await?;
```
dubek, line 141: (optimization)
The first time this is called it will surely fail, because the kbs_client has no cookies.
Maybe we can locally check if the cookie store is empty, and if it is, start directly from the `POST /auth` request.

dubek (commented):
Looks good. A few questions/comments below:

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -33,13 +35,35 @@ impl KbcInterface for Kbc {
         Err(anyhow!("Decrypt Payload API of this KBC is unimplemented."))
     }
 
-    async fn get_resource(&mut self, _description: &str) -> Result<Vec<u8>> {
-        Err(anyhow!("Unimplemented"))
+    #[allow(unused_assignments)]
+    async fn get_resource(&mut self, description: &str) -> Result<Vec<u8>> {
+        let desc: ResourceDescription = serde_json::from_str::<ResourceDescription>(description)?;
+        let mut resource_url = String::default();
+        let resource_type = desc
```
Xynnn007, line 42: I suggest to impl `TryInto<String>` for `ResourceDescription`. Then it will be convenient whenever it needs to convert a `ResourceDescription` into a `String`

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -33,13 +35,35 @@ impl KbcInterface for Kbc {
...
+            .optional
+            .get("type")
+            .ok_or_else(|| anyhow!("Invalid Resource description: Missing `type` filed"))?;
+        let resource_tag = desc
+            .optional
+            .get("tag")
+            .ok_or_else(|| anyhow!("Invalid Resource description: Missing `tag` filed"))?;
+        if let Some(resource_repo) = desc.optional.get("repository") {
```
Xynnn007, line 50: Seems that the kbs' url in `ResourceDescription` will overwrite the url of the `cc-kbc` here. Is the reverse more appropriate?

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +120,60 @@ impl Kbc {
             .ok_or_else(|| anyhow!("TEE rsa key missing"))?;
         response.decrypt_output(key)
     }
+
+    fn tee(&self) -> String {
+        self.tee.clone()
```
Xynnn007, line 125: I suggest the return value here be `&str`, then the caller can decide whether to `clone()` or not. This helps to decrease unnecessary memory copy.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +120,60 @@ impl Kbc {
...
+    }
+
+    fn kbs_uri(&self) -> String {
+        self.kbs_uri.clone()
```
Xynnn007, line 129: Same as mentioned, that the return value here be `&str` might be better.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +120,60 @@ impl Kbc {
...
+            }
+            res = self.kbs_client()?.get(&resource_url).send().await?;
+        }
+
+        if res.status() == reqwest::StatusCode::OK {
+            let response = res.json::<Response>().await?;
+            return Ok(response);
+        } else {
+            // TODO: Parse Error description payload
+            return Err(anyhow!("Request KBS resource failed"));
```
Xynnn007, line 176: We can replace all `Err(anyhow!(...))` with a convenient macro
```rust
bail!(...)
```

src/kbc_modules/cc_kbc/mod.rs
jialez0: Good suggestion, but it shouldn't be this PR's work.

src/kbc_modules/cc_kbc/mod.rs
jialez0: Done.

src/kbc_modules/cc_kbc/mod.rs
jialez0: Updated, thanks!

src/kbc_modules/cc_kbc/mod.rs
jialez0: This makes sense, but the cookie store in reqwest client is private, which is inconvenient. I use explicit cookie store to achieve this.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -33,13 +41,35 @@ impl KbcInterface for Kbc {
...
-    async fn get_resource(&mut self, _description: &str) -> Result<Vec<u8>> {
-        Err(anyhow!("Unimplemented"))
+    #[allow(unused_assignments)]
+    async fn get_resource(&mut self, description: &str) -> Result<Vec<u8>> {
+        let desc: ResourceDescription = serde_json::from_str::<ResourceDescription>(description)?;
+        let mut resource_url = String::default();
+        let resource_type = desc
+            .optional
+            .get("type")
+            .ok_or_else(|| anyhow!("Invalid Resource description: Missing `type` filed"))?;
```
dubek, line 51: typo: filed -> field
(also in line 55)

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +127,116 @@ impl Kbc {
...
+            // TODO: Parse Error description payload
+            bail!("Request KBS resource failed")
+        }
+    }
+}
+
+fn build_http_client() -> Result<reqwest::Client> {
+    reqwest::Client::builder()
+        .cookie_store(true)
+        .user_agent("attestation-agent-cc-kbc")
```
dubek, line 238: it would be helpeful (for logging) to add the AA version here, if we have it.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +127,116 @@ impl Kbc {
...
+            .http_client()
+            .get(&resource_url)
+            .header(
+                reqwest::header::COOKIE,
+                self.cookie_store.cookies(&parsed_kbs_uri).ok_or_else(|| {
+                    anyhow!("Established session again but cookie_store is still empty.")
+                })?,
+            )
+            .send()
+            .await?;
```
dubek, line 223: instead of repeating this part again, I suggest a loop, something like:

```rust
for attempt in 1..3 {
  log!("trying to get resource, attempt {attempt}");
  request resource
  if UNAUTHORIZED {
    establish_kbs_session
    sleep 5 seconds
  } else {
   break
  }
} 
```

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +127,116 @@ impl Kbc {
...
+        Ok(())
+    }
+
+    async fn request_kbs_resource(&mut self, resource_url: String) -> Result<Response> {
+        let parsed_kbs_uri =
+            Url::parse(self.kbs_uri()).map_err(|e| anyhow!("Parse KBS URL Failed: {:?}", e))?;
+
+        if self.cookie_store.cookies(&parsed_kbs_uri).is_none() {
+            self.establish_kbs_session().await?;
+        }
```
dubek, line 196: I know it was my suggestion to check the cookie store, but maybe instead of keeping our own cookies store and handling the set-cookie header manually, we can revert to the original implementation (with built-in private cookie store), and just keep a flag `bool is_authenticated`. It starts as false, and after successful `/attest` it is changed to true. If we ever get UNAUTHORIZED we changed it to false again.

Don't you think it's simpler than the whole handling of cookie_store ?

dubek (commented):
Few more comments:

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +127,86 @@ impl Kbc {
...
+                continue;
+            }
+
+            if res.status() == reqwest::StatusCode::OK {
+                let response = res.json::<Response>().await?;
+                return Ok(response);
+            } else {
+                // TODO: Parse Error description payload
+                bail!("Request KBS resource failed")
+            }
```
dubek, line 195: [style] consider changing these `if`s and `else` into a `match req.status() { ... }`.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -94,4 +127,86 @@ impl Kbc {
...
+        }
+
+        self.authenticated = true;
+
+        Ok(())
+    }
+
+    async fn request_kbs_resource(&mut self, resource_url: String) -> Result<Response> {
+        for attempt in 0..KBS_GET_RESOURCE_MAX_ATTEMPT {
+            log::info!("CC-KBC: trying to get resource, attempt {attempt}");
```
dubek, line 176: just for a clearer log message, maybe change the `for` statement to:

```suggestion
        for attempt in 1..=KBS_GET_RESOURCE_MAX_ATTEMPT {
            log::info!("CC-KBC: trying to get resource, attempt {attempt}");
```

dubek (approved):
Looks good to me.

Minor style comments below; can also wait for another PR.

src/kbc_modules/cc_kbc/mod.rs
jialez0: Done!

dubek (approved):

fitzthum (approved):
LGTM. 

It's possible that we might want to tweak how resources are handled a bit after a more general discussion, but this seems fine for now.

Xynnn007 (approved):
Great, lgtm!

Alex-Carter01:
Any reason for reqwest vs request?

PR #124, opened 2023-01-12
Support ttRPC feature option

### LINKED ISSUE #216: [Feature Request] ttrpc as RPC module
Now we use `gRPC` for AA app. As mentioned in https://github.com/kata-containers/kata-containers/issues/6027 we want to support ttrpc in AA.  At first glance, there might be two ways:
- Add a new binary probably named `ttrpc-bin`. This binary, rather than in `app` now, uses `ttrpc` as RPC server rather than `gRPC`.
- Add support directly for `app/` now, and provide a launch parameter to specify whether `ttrpc` or `gRPC` will be used.

Seems that the first one is better, for it does not depend on `gRPC` when compiled.

jialez0:
Fix: confidential-containers/guest-components#216 
@jiangliu @Xynnn007 @arronwy


app/src/main.rs
```diff
@@ -64,8 +86,83 @@ async fn main() -> Result<()> {
...
+        .register_service(keyprovider_service);
+    keyprovider_server.start().unwrap();
+
+    let mut getresource_server = Server::new()
+        .bind(&getresource_socket)
+        .unwrap()
+        .register_service(getresource_service);
+    getresource_server.start().unwrap();
+
+    loop {}
```
Xynnn007, line 154: Why we need a loop here?

Xynnn007:
Another question is do we need to check the feature `grpc` and `ttrpc` are exclusive? If both are enabled or neither is enabled, should compiler to report some errors? cc @jiangliu

Xynnn007:
So as suggested, a compile-time macro check like 
```rust
#[cfg(any(
    not(any(feature = "grpc", feature = "ttrpc")),
    all(feature = "grpc", feature = "ttrpc"),
))]
compile_error!(
    "One and exactly one feature of `grpc` or `ttrpc` must be enabled."
);
```

would help

app/src/main.rs
jialez0: Here we need to hold the main process, or it will close the program directly.

jialez0:
> So as suggested, a compile-time macro check like
> 
> ```rust
> #[cfg(any(
>     not(any(feature = "grpc", feature = "ttrpc")),
>     all(feature = "grpc", feature = "ttrpc"),
> ))]
> compile_error!(
>     "One and exactly one feature of `grpc` or `ttrpc` must be enabled."
> );
> ```
> 
> would help

Added, thanks!

app/src/rpc/mod.rs
```diff
@@ -0,0 +1,16 @@
...
+
+pub mod getresource;
+pub mod keyprovider;
+
+#[cfg(feature = "ttrpc")]
+use std::collections::HashMap;
+#[cfg(feature = "ttrpc")]
+pub mod ttrpc_protocol;
+#[cfg(feature = "ttrpc")]
+pub type TtrpcService = HashMap<String, Box<dyn ::ttrpc::MethodHandler + Send + Sync>>;
```
jiangliu, line 14: how about `std::collection::HashMap<String, Box<dyn ::ttrpc::MethodHandler + Send + Sync>>` to reduce one conditional compilation?

Makefile
```diff
@@ -20,11 +20,21 @@ ARCH ?= $(shell uname -m)
...
-    feature := --no-default-features --features
+    features += $(KBC)
+else
+    features += sample_kbc
+endif
+
+ifdef ttrpc
+    features += ttrpc
+else
+    features += grpc
```
jiangliu, line 37: Is it supported to enable both `ttrpc` and `grpc`?

app/Cargo.toml
```diff
@@ -6,9 +6,11 @@ publish = false
 edition = "2021"
 
 [dependencies]
-tokio = { version = "1.0", features = ["full"] }
-tonic = "0.7.2"
-prost = "0.10.4"
+tokio = { version = "1.0", features = ["full"], optional = true }
```
jiangliu, line 9: Is it possible to specify tokio features needed instead of `full`?

app/Cargo.toml
```diff
@@ -22,11 +24,14 @@ attestation_agent = { path = "../", default-features = false }
 async-trait = "0.1.56"
 
 [build-dependencies]
-tonic-build = "0.7.2"
+tonic-build = { version = "0.7.2", optional = true }
+ttrpc-codegen = { version = "0.4.1", optional = true }
 shadow-rs = "0.11.0"
```
jiangliu, line 29: Can you help to issue another PR to remove shadow-rs? It has been removed from ocicrypt-rs and image-rs.

app/Cargo.toml
```diff
@@ -22,11 +24,14 @@ attestation_agent = { path = "../", default-features = false }
...
 
 [features]
-default = ["sample_kbc"]
+default = ["sample_kbc", "grpc"]
+grpc = ["tokio", "tonic", "prost", "tonic-build"]
+ttrpc = ["dep:ttrpc", "ttrpc-codegen", "protobuf"]
```
jiangliu, line 34: Is it OK to change "dep:ttrpc" to "ttrpc"? It's not supported by older rust versions.

app/build.rs
```diff
@@ -3,9 +3,33 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+#[cfg(feature = "ttrpc")]
+use ttrpc_codegen::{Codegen, Customize, ProtobufCustomize};
```
jiangliu, line 7: This can be moved into the code body below at line 17 for simplicity.

app/src/main.rs
```diff
@@ -10,62 +10,35 @@ use anyhow::*;
...
 use log::*;
-use std::net::SocketAddr;
 use std::sync::Arc;
-use tokio::sync::Mutex;
 
-mod grpc;
+#[cfg(any(
+    not(any(feature = "grpc", feature = "ttrpc")),
+    all(feature = "grpc", feature = "ttrpc"),
+))]
```
jiangliu, line 18: Is it ok to enable both ttrpc and grpc?

app/src/rpc/getresource/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+use futures::executor::block_on;
+#[cfg(feature = "ttrpc")]
+use ttrpc::proto::{Code, Status};
+
+#[derive(Debug, Default)]
+pub struct GetResource {}
+
+#[cfg(feature = "grpc")]
+#[tonic::async_trait]
+impl GetResourceService for GetResource {
```
jiangliu, line 42: You can use a tricky to simplify the condition compilations:
```
#[cfg(feature = "grpc")]
mod grpc {
     use ...
     #[tonic::async_trait]
     impl GetResourceService for GetResource {
     }
}
```

app/Cargo.toml
jialez0: In the latest version of rust, if we do not add this prefix, the compiler will encounter the following error because the dependency name is the same as the feature name:

```shell
$ cargo build --release --no-default-features --features " sample_kbc ttrpc grpc" --target x86_64-unknown-linux-gnu
error: failed to parse manifest at `/root/data/jiale/ttrpc/attestation-agent/app/Cargo.toml`

Caused by:
  optional dependency `ttrpc` is not included in any feature
  Make sure that `dep:ttrpc` is included in one of features in the [features] table.
make: *** [Makefile:115: build] Error 101
```

app/Cargo.toml
jialez0: Removed!

app/Cargo.toml
jialez0: Done.

Makefile
jialez0: Supported it now, thanks.

app/src/ttrpc.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2021 Alibaba Cloud
```
jiangliu, line 1: 2023:）

app/src/ttrpc.rs
```diff
@@ -0,0 +1,90 @@
...
+    debug!(
+        "KeyProvider ttRPC service listening on: {:?}",
+        keyprovider_socket
+    );
+    debug!(
+        "GetResource ttRPC service listening on: {:?}",
+        getresource_socket
+    );
+
+    if !Path::new(AA_UNIX_SOCKET_DIR).exists() {
```
arronwy, line 57: We may need take socket dir from inputs.

arronwy (approved):
Thanks @jialez0 LGTM! Verified works with current `ocicrypt-rs`.

app/Cargo.toml
arronwy: Maybe we can add a prefix to feature name like `proto_ttrpc` and `proto_grpc` or some better.

Xynnn007:
Hi @jiangliu  is this pr ok or there is something else blocking?

jiazhang0 (approved):

PR #117, opened 2022-12-30
Compatible API for KBC

Xynnn007:
This PR aims to resolve the compatible problem for KBCes, including the following:
- Defined a common AnnotationPacket format for all KBCes.
- Brought decryption logic out of each KBC module
- Added documents and specs
- Added new feature openssl to support hardware acceleration for some HW

Warning: This pr will make current CI (image decryption) using sample-kbc, eaa-kbc not work. Thus we need to make new test images after the next Pr which will update the key provider module

PR #127, opened 2023-01-30
CC-KBC: Implement decrypt_payload method

### LINKED ISSUE #215: cc-kbc: Implement `decrypt_payload` interface.
Now CC-KBC is supported `get_resource` trait interface, but the `decrypt_payload` interface is still blank.

jialez0:
Resolve: confidential-containers/guest-components#215

### DIFF
--- a/sample_keyprovider/src/enc_mods/cc_kbc_enc/mod.rs
+++ b/sample_keyprovider/src/enc_mods/cc_kbc_enc/mod.rs
@@ -1,0 +1,49 @@
+// Copyright (c) 2023 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use aes_gcm::aead::{Aead, KeyInit};
+use aes_gcm::{AeadCore, Aes256Gcm, Key};
+use anyhow::*;
+use rand::rngs::OsRng;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize)]
+pub struct AnnotationPacket {
+    // Key Resource ID (URL)
+    // Format:
+    // `cc_kbc://127.0.0.1:8080/test_repo/key/id_1`
+    pub kid: String,
+    // Encrypted key to unwrap (base64-encoded)
+    pub wrapped_data: String,
+    // Initialisation vector (base64-encoded)
+    pub iv: String,
+    // Wrap type to specify encryption algorithm and mode
+    pub wrap_type: String,
+}
+
+pub fn enc_optsdata_gen_anno(optsdata: &[u8], params: Vec<String>) -> Result<String> {
+    let (key_file_path, kid) = params[0].split_once(':').ok_or(anyhow!(
+        "Failed to parse parameters: {:?}, need key file path and key URL in KBS split by ':'",
+        params
+    ))?;
+
+    let key = std::fs::read(key_file_path).map_err(|e| anyhow!("Read Key file failed: {}", e))?;
+    let aes_key = Key::<Aes256Gcm>::from_slice(&key);
+    let nonce = Aes256Gcm::generate_nonce(&mut OsRng);
+
+    let cipher = Aes256Gcm::new(aes_key);
+    let encrypt_optsdata = cipher
+        .encrypt(&nonce, optsdata)
+        .map_err(|e| anyhow!("Eecrypt failed: {:?}", e))?;
+
+    let annotation = AnnotationPacket {
+        kid: kid.to_string(),
+        wrapped_data: base64::encode(encrypt_optsdata),
+        iv: base64::encode(nonce.to_vec()),
+        wrap_type: "A256GCM".to_string(),
+    };
+
+    serde_json::to_string(&annotation).map_err(|_| anyhow!("Serialize annotation failed"))
+}
--- a/src/kbc_modules/cc_kbc/crypto/mod.rs
+++ b/src/kbc_modules/cc_kbc/crypto/mod.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+use super::KBS_URL_PREFIX;
 use crate::kbc_modules::cc_kbc::kbs_protocol::message::Response;
 use aes_gcm::aead::{Aead, KeyInit};
 use aes_gcm::Aes256Gcm;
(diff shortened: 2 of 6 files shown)


src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -39,8 +39,30 @@ impl KbcInterface for Kbc {
...
+        let key = self.decrypt_response_output(response)?;
+
+        let iv = base64::decode(annotation_packet.iv)
+            .map_err(|e| anyhow!("Failed to decode initialization vector: {}", e))?;
+        let wrapped_data = base64::decode(annotation_packet.wrapped_data)
+            .map_err(|e| anyhow!("Failed to decode wrapped key: {}", e))?;
+
+        match annotation_packet.wrap_type.as_str() {
+            crypto::AES_GCM_256_ALGORITHM => crypto::aes_gcm_256_decrypt(&wrapped_data, &key, &iv)
+                .map_err(|e| anyhow!("Failed to decrypt annotation: {}", e)),
```
dubek, line 61: Here we use AES-GCM-256 (which I think is good) but in online_sev_kbc we use openssl's AES-256-CTR:

https://github.com/confidential-containers/attestation-agent/blob/main/src/kbc_modules/online_sev_kbc/mod.rs#L60-L69

@fitzthum : I understand that offline_fs_kbc is used on s390x and it will gain from OpenSSL's optimizations. Why do we need OpenSSL and AES-256-CTR in the other KBCs?

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -17,7 +17,20 @@ const RSA_ALGORITHM: &str = "RSA1_5";
 const RSA_PUBKEY_LENGTH: usize = 2048;
 const NEW_PADDING: fn() -> PaddingScheme = PaddingScheme::new_pkcs1v15_encrypt;
 
-const AES_GCM_256_ALGORITHM: &str = "A256GCM";
+pub const AES_GCM_256_ALGORITHM: &str = "A256GCM";
```
dubek, line 20: while we're at it, I think the correct name should be `AES_256_GCM_ALGORITHM` (similarly the function name `aes_256_gcm_decrypt` and in line 137).

dubek (commented):
Thanks, this looks good. A few small comment and a question (about other KBCs...):

src/kbc_modules/cc_kbc/crypto/mod.rs
```diff
@@ -17,7 +17,20 @@ const RSA_ALGORITHM: &str = "RSA1_5";
...
+
+#[derive(Serialize, Deserialize)]
+pub struct AnnotationPacket {
+    // Key Resource URL
+    // e.g: http://127.0.0.1:8080/kbs/v0/resource/test_repo/key/id_1
+    pub key_url: String,
```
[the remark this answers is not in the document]
jialez0: I have corrected this. The `kid` here should actually be a new unified format proposed in this issue: https://github.com/confidential-containers/documentation/issues/85. This format will be used in CoCo to index all resources that need to be obtained from AA. The `kid` here is consistent with it. The format is:

```
cc_kbc://127.0.0.1:8080/test_repo/key/id_1
```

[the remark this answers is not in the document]
Xynnn007: Hi @dubek , I'm going to bring the logic of parsing resource URL out of logic of KBCs. For example, 
- `cc_kbc://127.0.0.1:8080/test_repo/key/id_1`. AA will dispatch this url to `cckbc`, and let `cckbc` determine how to process `127.0.0.1:8080/test_repo/key/id_1`. In this scenario, it will convert this into a http link `https://127.0.0.1:8080//kbs/v0/resources/key/id_1` and make a request in kbs protocol
- `offlinefskbc://key/1`. AA will dispatch this url to `offline fs kbc`, and let `offline fs kbc` determine how to process `key/1`. Based on the mechinism of offline kbc, it will look up the resource.json to directly return.

The PR is under development https://github.com/confidential-containers/attestation-agent/pull/117
The mechinism is described in https://github.com/confidential-containers/documentation/issues/85

src/kbc_modules/cc_kbc/mod.rs
Xynnn007: There is something I'm working on that might be related. As openssl is good for performance on s390x, I made [two features of crypto impls](https://github.com/confidential-containers/attestation-agent/pull/117/files#diff-899d7ab1242a214d596c21f0dda8d4df6696ba8186d7d931f253359d0ef0b78aR18) (purely rust and openssl), which can be selected by user when compiles AA.

I also found that [openssl supports AES-GCM](https://github.com/confidential-containers/attestation-agent/pull/117/files#diff-58cd7a44a49dd41a612e3eb7c58f89c41ae041b6604105abc2b3481f89f13d4eR39), but a different api from AES-CTR. I wonder whether we can use AES-GCM in offline-kbc?

fitzthum: I think `online_sev_kbc` just uses the same cryto stuff as the offline KBC out of convenience. There has  been some discussion of unifying all of this so that container images are truly interchangeable, which I think would be a good idea.

jialez0:
@dubek Thanks for reviewing! We can try to get this PR merged first, so @Xynnn007 can rebase confidential-containers/attestation-agent#117 then some defects and confusion can be solved in that PR.. @fitzthum

sample_keyprovider/src/enc_mods/cc_kbc_enc/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+pub fn enc_optsdata_gen_anno(optsdata: &[u8], params: Vec<String>) -> Result<String> {
+    let (key_file_path, kid) = params[0].split_once(':').ok_or(anyhow!(
+        "Failed to parse parameters: {:?}, need key file path and key URL in KBS split by ':'",
+        params
+    ))?;
+
+    let key = std::fs::read(key_file_path).map_err(|e| anyhow!("Read Key file failed: {}", e))?;
+
+    let mut iv: [u8; 12] = [0; 12];
```
dubek, line 45: I think that the number `12` can be replaced by `Aes256Gcm::NonceSize` (haven't tested).

dubek (approved):
Looks good to me.
Few minor optional comments below.

sample_keyprovider/src/enc_mods/cc_kbc_enc/mod.rs
jialez0: I found that `aes-gcm` crate does not provide such a constant. The `NonceSize` provided by it is an `ArrayLength<u8>` trait: https://docs.rs/aes-gcm/0.10.1/aes_gcm/type.Nonce.html?search=NonceSize. Therefore, it cannot be used as use here. Perhaps it is more appropriate to use a constant defined by ourselves like:
```
const AES_256_GCM_NONCE_SIZE: usize = 12;
```

Xynnn007 (approved):

jialez0:
@fitzthum

sample_keyprovider/src/enc_mods/cc_kbc_enc/mod.rs
dubek: Yes, you're right. The constant you added is good.

----

My side research:

I found that if you add:
```
use aes_gcm::AeadCore;
```
then you get a generate_nonce function:
```
let nonce = Aes256Gcm::generate_nonce(OsRng);
```
which must return the correct length for Aes256Gcm. Then we can just pass this Aes256Gcm::Nonce variable to the encrypt function (instead of `iv: &[u8]`) -- but I'm actually having trouble finding the correct type for such an argument.

jialez0: That's great. The return type of `Aes256Gcm::generate_nonce` is `Aes256Gcm::Nonce`, which is actually `GenericArray<u8, NonceSize>`. We can use `to_vec()` to convert this struct to `Vec<u8>`.

fitzthum (approved):
LGTM. There are still some things to figure out regarding KBC compatibility, but we can adjust later.

PR #137, opened 2023-02-20
CC-KBC Fix: Add default value for `get_resource` param.

jialez0:
In order to be compatible with the current simplest resource request scheme, when the optional field in the resource description is empty, we need to use the default field to fill in, instead of throwing an error.


Xynnn007 (approved):
Great. As a temporary fix it can work for CC-KBC. After KBS URI is brought in, the fields can be parsed from the resource's uri.

fitzthum (approved):

PR #139, opened 2023-02-23
Sample-keyprovider: fix the wrap-type name of cc-kbc-enc

jialez0:
The purpose is to keep consistent with the decryption module in AA. (https://github.com/confidential-containers/attestation-agent/blob/main/src/common/crypto/mod.rs#L19)


Xynnn007:
Now the cut of v0.4.0 will not include the code you've given. As wrap type like `A256GCM` follows RFC7518, but `aes-gcm-256` doesn't. Maybe we can wait after v0.4.0 is released, unify them to RFC7518 and then make new test images.

jialez0:
> Now the cut of v0.4.0 will not include the code you've given. As wrap type like A256GCM follows RFC7518, but aes-gcm-256 doesn't. Maybe we can wait after v0.4.0 is released, unify them to RFC7518 and then make new test images.

A related fix is https://github.com/confidential-containers/attestation-agent/commit/d4dd4cfb5568fa1a60a0930974365386654b535a#diff-6fe33f515a2cbb8fb95a6e4e543c91df6ef7dc142e57d401bc88f87973da5e53R92

@Xynnn007 OK, I will close this PR, thanks!

PR #136, opened 2023-02-20
CC-KBC | SGX Occlum Attester

Xynnn007:
> **warning** CI machine is not prepared, related scripts and yamls will not work

PR #141, opened 2023-02-27
CC-KBC | Provide compilation features for different TEE attesters

### LINKED ISSUE #212: CC-KBC | Provide compilation features for different TEE attesters
Now CC-KBC compiles all the TEE attesters it supports by default, which is redundant. We need to set different features to specify which TEE attester the compiled CC-KBC contains.

cc @fitzthum @Xynnn007

jialez0:
Resolve: confidential-containers/guest-components#212


jialez0:
cc @fitzthum

.github/workflows/cc_kbc.yml
```diff
@@ -47,6 +47,7 @@ jobs:
       - name: Build and install with cc_kbc feature
         run: |
           make KBC=cc_kbc && make install
+          make KBC=cc_kbc_tdx && make install
```
dubek, line 50: The new `make` invocation doesn't fit the step's `name`.
Also below on the clippy step we don't use the `cc_kbc_tdx` feature.

Suggestsion: Consider adding a matrix dimension above (line 22):

```yaml
      matrix:
        rust:
          - stable
        kbc:
          - cc_kbc
          - cc_kbc_tdx
```

and then use `${{ matrix.kbc }}` in each of the steps to choose the current one.

Xynnn007 (commented):
Great. Some same work is involved in another PR https://github.com/confidential-containers/attestation-agent/pull/136/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R52 but it's good to me to rebase after this PR

A little suggestion: probably we can also add rust fmt check in the cc-kbc CI for different attesters in the way that @dubek suggested?

jialez0:
> A little suggestion: probably we can also add rust fmt check in the cc-kbc CI for different attesters in the way that @dubek suggested?

@Xynnn007 In `basic.yml`, we use `cargo fmt --all -- --check`, this checked all features of AA, so in `cc_kbc.yml`, we don't need to check the format again.

.github/workflows/cc_kbc.yml
```diff
@@ -46,11 +49,11 @@ jobs:
 
       - name: Build and install with cc_kbc feature
```
dubek, line 50: maybe change this to:

```
name: Build and install with ${{ matrix.kbc }} feature
```

(assuming that you can use these interpolations in the `name` field; not sure)

dubek (approved):
Looks good. Small comment below:

fitzthum (approved):
LGTM. Ideally we will build for all platforms in production. I will rebase my SNP attester to use.

PR #129, opened 2023-02-06
KBS URI

Xynnn007:
As https://github.com/confidential-containers/guest-components/issues/191, this PR includes the following
- Introduce KBS URI scheme
- implement cc-kbc's API
- implement sample-kbc's API
- implement offline-fs-kbc's API
- implement offline-sev-kbc's API
- implement online-sev-kbc's API
- implement eaa-kbc's API
- document about the KBS URI

Please notice that all the `<kbs-addr>` field in a kbs uri now will be ignored, all kbc will use its own member `kbs_url` instead.

Also, this PR modifies the gRPC API of `get_resource`, as duplicated information is carried in the kbs uri

PR #143, opened 2023-02-28
CoCo Keyprovider

Xynnn007:
**warning**: This PR dependes on https://github.com/confidential-containers/attestation-agent/pull/129, which brings kbs resource uri scheme. I have test locally using AA based on that PR.

#### What's new
- Combined sample key provider, offline-fs-kbs and cc_kbc_enc, as now all different KBC share a same format of AnnotationPacket
- Added function of exporting key for the script `generate-keys`
- Defined input parameters for skopeo to use for the CoCo Keyprovider
- Gave a guide to test encrypting & decrypting image


sample_keyprovider/README.md
```diff
@@ -32,24 +73,42 @@ vim ocicrypt.conf:
 Copy the image you want to encrypt to your current directory. This example uses a *busybox* image:
 
 ```
-skopeo copy docker://busybox oci:busybox
+OCICRYPT_KEYPROVIDER_CONFIG=ocicrypt.conf skopeo copy --encryption-key provider:attestation-agent:<parameters> docker://busybox oci:busybox_encrypted
 ```
+After encryption, it can be seen that busybox-encrypted is generated in the current directory.
 
-Encrypt the container image. You can pass in the special parameters required by the encryption module through the command line parameters of skopeo (here, the string "test" is taken as an exemplary special parameter): 
+Or we can directly push the encrypted image to the remote iamge registry
```
jialez0, line 80: Nit: `iamge` -> `image`.

sample_keyprovider/README.md
```diff
@@ -1,20 +1,61 @@
...
+The following guide will help make an encrypted image using CoCo keyprovider and decrypt that.
 
-## Test KBC modules
+## Related tools
 
-### Encryption
+Let's firstly introduce the related tools to help generate keys.
 
-Build and run sample keyprovider. You must explicitly specify the encryption module to use through the --features parameter (sample_enc for example): 
+### Test Key Generation and Offline-Fs-KBC
```
jialez0, line 14: This chapter is specific to Offline-Fs-KBC, so it is recommended to put it at the end and describe how to test Offline-Fs-KBC in a separate chapter.

The main body of README should guide the user to directly generate an independent key.

sample_keyprovider/README.md
```diff
@@ -1,20 +1,61 @@
...
+
+`tools/generate_keys.sh` is a helper script to generate ten different base64-encoded KEK. Also, it can help to generate the `aa-offline_fs_kbc-keys.json` used by offline-fs-kbc.
+**Warning**: `jq` is required, so please install jq before use the script. For example on ubuntu, `apt install jq` could help.
+
+#### Generate `aa-offline_fs_kbc-keys.json`
```
jialez0, line 19: As above, this chapter is specific to Offline-Fs-KBC.

sample_keyprovider/README.md
```diff
@@ -1,20 +1,61 @@
...
+  "test-repo/key/key_id8": "eXIlv83nTjfyeLZcCvTda9ypYIYj83eGjbqjoVFAQHA="
+}
+```
+
+#### Use the Generated Key to Encrypt Image
+
+If we want to use key of id `test-repo/key/key_id1` in `aa-offline_fs_kbc-keys.json` to encrypt an image, we firstly export the key content to file `key1`
+
+```
+tools/generate_keys.sh export aa-offline_fs_kbc-keys.json test-repo/key/key_id1 key1
```
jialez0, line 46: Can we not pass `aa-offline_fs_kbc-keys.json`, but directly use the tool to generate a key?

sample_keyprovider/README.md
```diff
@@ -32,24 +73,42 @@ vim ocicrypt.conf:
...
+The `<parameters>` in skopeo command is a key-value list separated by double colons. Here are the defined keys:
+- `sample`: Not required. Either `true` or `false`. If not set, use `false`. This value indicates whether the hardcoded encryption key is used. This works the same way as `sample keyprovider`.
+- `keyid`: Required if `sample` is not enabled. It is a KBS Resource URI, s.t. `kbs://<kbs-addr>/<repo>/<type>/<tag>`. When decryption occurs, the `keyid` value is used to index the KEK.
+- `keypath`: Required if `sample` is not enabled. A local filesystem path, absolute path recommended. Specify the KEK to encrypted the image in local filesystem. KEK will be read from fs and then used to encrypt the image. This key's length must be 32 bytes.
+- `algorithm`: Not required. Indicate the encryption algorithm used. Either `A256GCM` or `A256CTR`. If not provided, use `A256GCM` by default as it is AEAD scheme.
 
-### Decryption
+### Examples
+
+- Use key `key1`, and keyid `kbs://example.org/test-repo/key/key_id1` to encrypt an image. In this way sample is disabled, and will use A256GCM (AES-256-GCM).
```
fitzthum, line 96: These parameters seem inconsistent. It says use "key `key1`", but there is no parameter called key. Is that supposed to be `keypath`? Shouldn't the keypath be something like `aa-offline_fs_kbc-keys.json`?

sample_keyprovider/README.md
```diff
@@ -32,24 +73,42 @@ vim ocicrypt.conf:
...
+```
+OCICRYPT_KEYPROVIDER_CONFIG=ocicrypt.conf skopeo copy --insecure-policy --encryption-key provider:attestation-agent:keypath=key1::keyid=kbs://example.org/test-repo/key/key_id1::algorithm=A256GCM docker://busybox oci:busybox:encrypted
+```
+
+- Use sample key provider to encrypt an image
+```
+OCICRYPT_KEYPROVIDER_CONFIG=ocicrypt.conf skopeo copy --insecure-policy --encryption-key provider:attestation-agent:sample=true docker://busybox oci:busybox:encrypted
+```
```
fitzthum, line 104: Can we add something about how to test inspect if an image is actually encrypted. Skopeo usually does not warn if the encryption fails.

sample_keyprovider/src/enc_mods/crypto.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2023 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use aes_gcm::{aead::Aead, aes::Aes256, Aes256Gcm, Key, Nonce};
+use anyhow::*;
+use strum_macros::EnumString;
+
+pub const HARDCODED_KEY: &[u8] = &[
```
fitzthum, line 10: Maybe add a comment saying that this is just for the sample.

sample_keyprovider/tools/generate_keys.sh
```diff
@@ -0,0 +1,66 @@
...
+		- ./generate_keys.sh export <path-to-aa-offline_fs_kbc-keys.json> <key-uri> <output-path>
+			Export the key of <key-uri> in the given "aa-offline_fs_kbc-keys.json". The result
+			key will be exported to <output-path>
+EOT
+    exit
+}
+
+create_keys() {
+	for i in $(seq 1 "$1"); do
+		keys[test-repo/key/key_id$i]=$(head -c32 < /dev/random | base64)
```
fitzthum, line 22: Maybe there should be an option to set the repo that will be used.

fitzthum (commented):
Looks good in general, but a few comments.

sample_keyprovider/README.md
Xynnn007: Yes. I will give two ways to ensure. one is to check the manifest and the other is to use the offline-fs-kbc to decrypt.

> Skopeo usually does not warn if the encryption fails.

FYI. I once meet a situation where an image is encrypted but could be directly run by `nerdctl`. Finally the reason is that `nerdctl` will call keyprovider grpc using the config under `/etc/containerd/ocicrypt`. Sometimes similiar tools 
may cause the illusion that the image is not encrypted, but in fact the logic behind it calls the grpc of the keyprovider to decrypt the image

sample_keyprovider/src/enc_mods/crypto.rs
Xynnn007: Great, I'll add that

sample_keyprovider/tools/generate_keys.sh
Xynnn007: Yes, added

sample_keyprovider/README.md
Xynnn007: Yes, reordered

sample_keyprovider/README.md
Xynnn007: We can generate a random key using `/dev/random` now. As the feature, I think when resource registration is prepared in KBS, we can integrate this function together with kbs, s.t.
- randomly generate a key
- use the key to encrypt image
- register the key into kbs

Xynnn007: If not in hurry, we can impl that in another PR when all basic functions are ok

sample_keyprovider/README.md
```diff
@@ -1,20 +1,61 @@
...
+
+#### Use the Generated Key to Encrypt Image
+
+If we want to use key of id `test-repo/key/key_id1` in `aa-offline_fs_kbc-keys.json` to encrypt an image, we firstly export the key content to file `key1`
+
+```
+tools/generate_keys.sh export aa-offline_fs_kbc-keys.json test-repo/key/key_id1 key1
+```
+
+Attention: When an image is being decrypted in offline-fs-kbc, we ignore the `<kbs-addr>` field of the key id field, thus in `aa-offline_fs_kbc-keys.json` we do not include the prefix `kbs://<kbs-addr>`.
```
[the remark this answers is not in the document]
Xynnn007: Yes, thanks. fixed

jialez0 (approved):
LGTM.

jialez0:
cc @fitzthum

Xynnn007:
> It is handy to specify the key ahead of time so that we can reuse keys for test images, but I agree it might be nice to have a feature where the keyprovider will generate a random key.

@fitzthum Hi Tobin, do you mean that we can just input the `keyid` to skopeo and then the encrypted image will be generated? I think it is pretty nice and hope so. The only question where the plaintext of the generated key should be output? This key is the generated key to encrypt, and which should be kept or input to KBS to wait for decryption requests.

The most straightforward way is to output the plaintext of the key to current dir, s.t. we can only use
```
OCICRYPT_KEYPROVIDER_CONFIG=ocicrypt.conf skopeo copy --insecure-policy --encryption-key provider:attestation-agent:keyid=default/key/1 docker://busybox oci:busybox:encrypted
```
And when the encryption is done, the key will be output to `./key`

A probably better way is to integrate this module into KBS, by which we can only use
```
OCICRYPT_KEYPROVIDER_CONFIG=ocicrypt.conf skopeo copy --insecure-policy --encryption-key provider:attestation-agent docker://busybox oci:busybox:encrypted
```

without any context, the module will automatically generate a key, encrypt and register the key into KBS (in a default repo with a random id)

The two ways I think should be a separate PR.

sample_keyprovider/README.md
fitzthum: Yeah, I agree we should try to create a seamless flow for encrypted and provisioning images.

fitzthum (approved):
We can still improve how the keys are specified, but this is good.

PR #144, opened 2023-03-02
Update tdx-attest-rs crate for buildgen build fix

jepio:
This pulls in the following fix which affects building the tdx-attest-sys crates when libclang is avilable but clang is not: https://github.com/intel/SGXDataCenterAttestationPrimitives/pull/28


jialez0 (approved):
Thanks! @jepio

PR #148, opened 2023-03-08
fix: use default repo and kbs-addr field

Xynnn007:
As <kbs-addr> field is not used in current kbs resource uri, it will be left empty. Thus a kbs resource uri will be from `kbs://<kbs-addr>/<repo>/<type>/<tag>` to `kbs:///<repo>/<type>/<tag>`

Also, we use "default" for the default repo name instead of "test-repo". In this way it will be more clearer and not confusing.

We can use these definitions in `image-rs`

cc @fitzthum @jialez0


jialez0 (approved):
LGTM. Thanks a lot! @Xynnn007 

@fitzthum Does this looks good to you?

fitzthum (approved):
Thanks

PR #150, opened 2023-03-12
fix: getresource with wrong kbs uri

Xynnn007:
The port field will be ignored when parse a kbs uri before. This will make the kbc fail to connect to a kbs. As it only receives `ip` rather than `ip:port` socket address.

### DIFF
--- a/src/kbc_modules/uri.rs
+++ b/src/kbc_modules/uri.rs
@@ -14,6 +14,10 @@ const RESOURCE_ID_ERROR_INFO: &str =
 
 const SCHEME: &str = "kbs";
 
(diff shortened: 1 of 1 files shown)


jialez0 (approved):

Xynnn007:
By default we use `kbs:///...` now and the uri here does not have a `kbs-addr` field. But in `get_resource` API, `image-rs` will automatically fill `kbs-addr` field and send to attestation-agent https://github.com/confidential-containers/image-rs/blob/main/src/resource/kbs/mod.rs#L152. Although this process is transparent to the user, attestation-agent should also parse the `port` field of the resource uri, which is ignored before. As a bug, this PR fixes to include the `port` field if a resource uri has.

PR #158, opened 2023-03-20
Bug-fix: Add KBS scheme prefix to URI

jialez0:


Xynnn007 (approved):
LGTM

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -146,7 +147,7 @@ impl Kbc {
 
         let challenge = self
             .http_client()
-            .post(format!("{kbs_uri}/{KBS_URL_PREFIX}/auth"))
+            .post(format!("{KBS_SCHEME}{kbs_uri}/{KBS_URL_PREFIX}/auth"))
```
sameo, line 150: So a URI is supposed to include a scheme. What happens if I create my KBC with `https://my-kbs.com` as the URI argument?
Shouldnt we assume that what is passed to the KBC constructor is a valid URI and e.g. validate that with the [url](https://crates.io/crates/url) crate?

jialez0: Ah, yes, you are right. I forgot this point. We should specify a scheme in the incoming parameters. Sorry!

PR #162, opened 2023-03-23
Fix cckbc uri scheme

Xynnn007:
When doing integration test, some bugs occurs and this pr can fix.
As the `reqwest` crate will not automatically try `http` when `https` fails, we must specify the KBS address by `aa_params`. That parameter will be delivered from the gRPC request.

### DIFF
--- a/protos/getresource.proto
+++ b/protos/getresource.proto
@@ -3,8 +3,9 @@ syntax = "proto3";
 package getresource;
 
 message GetResourceRequest {
-    string ResourceUri = 1;
+    string ResourcePath = 1;
     string KbcName = 2;
+    string KbsUri = 3;
 }
 
 message GetResourceResponse {
--- a/app/Cargo.toml
+++ b/app/Cargo.toml
@@ -30,7 +30,7 @@ tonic-build = { version = "0.7.2", optional = true }
 ttrpc-codegen = { version = "0.4.1", optional = true }
 
 [features]
-default = ["sample_kbc", "grpc", "ttrpc"]
+default = ["sample_kbc", "ttrpc"]
 grpc = ["tokio", "tonic", "prost", "tonic-build"]
 ttrpc = ["dep:ttrpc", "ttrpc-codegen", "protobuf"]
 sample_kbc = ["attestation_agent/sample_kbc"]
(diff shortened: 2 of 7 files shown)


protos/getresource.proto
```diff
@@ -5,6 +5,7 @@ package getresource;
 message GetResourceRequest {
     string ResourceUri = 1;
     string KbcName = 2;
+    string KbsUri = 3;
```
sameo, line 8: So what is the `KbsUri` going to look like compared to the `ResourceUri`?

Xynnn007: A resource uri is like `kbs://<uri-of-kbs>/<repo>/<resource>/<tag>` following KBS Resource URI for locating the resource. `KbsUri` is like `http://...` or `https://...` which is for pointing a remote KBS. Before we use the scheme `http://` to directly convert a uri `kbs://<uri-of-kbs>/<repo>/<resource>/<tag>` into `http://<uri-of-kbs>`. However, as the distinguishment of `http` and `https`, we need to specify the address of KBS which includes a `http` or a `https` protocol name.

Another choice is to let cc-kbc try to connect with `https` by default. If fails then downgrates to `http`. However `reqwest` crate does not support the automatical downgration, we need to manually implement this.

sameo: So `KbsUri` is a URI that contains a scheme and a host path. There's a bit of duplication between the two strings:

`kbs://example.cckbs.org:8081/alice/decryption-key/1` and `https://example.cckbs.org:8081/`.

Why not defining the API as:

```proto3
message GetResourceRequest {
    string ResourcePath = 1; // e.g. /alice/decryption-key/1
    string KbcName = 2; // e.g. cc_kbc
    string KbsUri = 3; // e.g. https://example.cckbs.org:8081/
```

With that approach there is no duplication in the arguments, and we have all the needed information to build fully qualified KBS compliant URIs.

Xynnn007: Oh, yes! Great suggestion! Let me fix this

sameo (approved):
Thanks @Xynnn007 !

jialez0 (approved):

PR #168, opened 2023-03-24
cc_kbc: Fix KBS URL handling

sameo:
The `cc_kbc` structure now carries the whole KBS URL, including scheme, host and port values.
We compare the KBS host value with the resource host address and bail if they don't match.

### DIFF
--- a/src/kbc_modules/cc_kbc/mod.rs
+++ b/src/kbc_modules/cc_kbc/mod.rs
@@ -19,7 +19,7 @@ use core::time::Duration;
 use crypto::{hash_chunks, TeeKey};
 use kbs_protocol::message::*;
 use kbs_types::ErrorInformation;
-use log::warn;
+use url::Url;
 use zeroize::Zeroizing;
 
 use super::{uri::ResourceUri, AnnotationPacket};
@@ -31,7 +31,7 @@ pub const KBS_URL_PREFIX: &str = "kbs/v0";
 
 pub struct Kbc {
     tee: String,
-    kbs_uri: String,
+    kbs_uri: Url,
     token: Option<String>,
     nonce: String,
     tee_key: Option<TeeKey>,
@@ -47,7 +47,7 @@ impl KbcInterface for Kbc {
     }
 
     async fn decrypt_payload(&mut self, annotation_packet: AnnotationPacket) -> Result<Vec<u8>> {
-        let key_url = self.resource_web_addr(&annotation_packet.kid)?;
+        let key_url = self.resource_to_kbs_uri(&annotation_packet.kid)?;
 
         let response = self.request_kbs_resource(key_url).await?;
         let key = Zeroizing::new(self.decrypt_response_output(response)?);
@@ -62,8 +62,7 @@ impl KbcInterface for Kbc {
 
     #[allow(unused_assignments)]
     async fn get_resource(&mut self, desc: ResourceUri) -> Result<Vec<u8>> {
-        let resource_url = self.resource_web_addr(&desc)?;
-
+        let resource_url = self.resource_to_kbs_uri(&desc)?;
         let response = self.request_kbs_resource(resource_url).await?;
 
         self.decrypt_response_output(response)
@@ -71,23 +70,29 @@ impl KbcInterface for Kbc {
 }
 
 impl Kbc {
-    pub fn new(kbs_uri: String) -> Kbc {
+    pub fn new(kbs_uri: String) -> Result<Kbc> {
+        // Check the KBS URI validity
+        let url = Url::parse(&kbs_uri).map_err(|e| anyhow!("Invalid URI {kbs_uri}: {e}"))?;
+        if !url.has_host() {
+            bail!("{kbs_uri} is missing a host");
+        }
+
         // Detect TEE type of the current platform.
         let tee_type = detect_tee_type();
 
         // Create attester instance.
         let attester = tee_type.to_attester().ok();
 
-        Kbc {
+        Ok(Kbc {
             tee: tee_type.to_string(),
-            kbs_uri,
+            kbs_uri: url,
             token: None,
             nonce: String::default(),
             tee_key: TeeKey::new().ok(),
             attester,
             http_client: build_http_client().unwrap(),
             authenticated: false,
-        }
+        })
     }
 
     fn generate_evidence(&self) -> Result<Attestation> {
@@ -134,7 +139,7 @@ impl Kbc {
     }
 
     fn kbs_uri(&self) -> &str {
-        &self.kbs_uri
+        self.kbs_uri.as_str()
     }
 
     fn http_client(&mut self) -> &mut reqwest::Client {
@@ -146,7 +151,7 @@ impl Kbc {
 
         let challenge = self
             .http_client()
-            .post(format!("{kbs_uri}/{KBS_URL_PREFIX}/auth"))
+            .post(format!("{kbs_uri}{KBS_URL_PREFIX}/auth"))
             .header("Content-Type", "application/json")
             .json(&Request::new(self.tee().to_string()))
             .send()
@@ -157,7 +162,7 @@ impl Kbc {
 
         let attest_response = self
             .http_client()
-            .post(format!("{kbs_uri}/{KBS_URL_PREFIX}/attest"))
+            .post(format!("{kbs_uri}{KBS_URL_PREFIX}/attest"))
             .header("Content-Type", "application/json")
             .json(&self.generate_evidence()?)
             .send()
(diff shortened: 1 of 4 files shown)


src/kbc_modules/uri.rs
```diff
@@ -79,21 +79,31 @@ impl From<ResourceUri> for url::Url {
...
+            Ok(url) => url
+                .host_str()
+                .ok_or_else(|| anyhow!("Invalid URL: {url}"))?
+                .to_string(),
+            Err(_) => kbs_uri.to_string(),
+        };
+
         if !resource_path.starts_with('/') {
-            bail!("resource_path should starts with '/'")
+            bail!("Resource path {resource_path} must starts with '/'")
```
mkulke, line 92:
```suggestion
            bail!("Resource path {resource_path} must start with '/'")
```

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -215,23 +220,25 @@ impl Kbc {
...
-    /// This function will **ignore** the kbs address the kid carries,
-    /// instead overwrite with the kbs_uri the [`Kbc`] carries.
-    /// Related issue: <https://github.com/confidential-containers/attestation-agent/issues/130>
-    pub fn resource_web_addr(&self, kid: &ResourceUri) -> Result<String> {
-        if self.kbs_uri != kid.kbs_addr {
-            warn!(
-                "The KBS address contained in the URI is not the same as the KBC's, use the KBC's.",
-            )
+    /// Convert a [`ResourceUri`] to a KBS URL.
+    pub fn resource_to_kbs_uri(&self, resource: &ResourceUri) -> Result<String> {
```
mkulke, line 224: do we want to have a unit test for the conversion?

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -215,23 +220,25 @@ impl Kbc {
...
+        let kbs_host = self
+            .kbs_uri
+            .host_str()
+            .ok_or_else(|| anyhow!("Invalid URL: {}", self.kbs_uri))?;
+        if kbs_host != resource.kbs_addr {
+            bail!(
```
jialez0, line 230: Currently, we cannot directly throw an error when the two are different, because the kbs address in the resource URI we currently use is actually omitted——It is a null value, which means that the resource URI entered by the current caller of AA for AA is like this: `kbs:///repo/type/tag`.

mkulke (commented):
I had to apply a small change (// twice in resource url"), but otherwise it worked for me

src/kbc_modules/cc_kbc/mod.rs
sameo: When do we use `cc_kbc` without a KBS address?

jialez0: When decrypting image annotation, the `kid` field which in Resource URI format included in the annotation is like this: `kbs:///a/b/c`.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -215,23 +220,25 @@ impl Kbc {
...
-        let repo = &kid.repository;
-        let r#type = &kid.r#type;
-        let tag = &kid.tag;
+        let kbs_addr = &self.kbs_uri();
+        let repo = &resource.repository;
+        let r#type = &resource.r#type;
+        let tag = &resource.tag;
         Ok(format!(
-            "http://{kbs_addr}/{KBS_URL_PREFIX}/resource/{repo}/{type}/{tag}"
+            "{kbs_addr}/{KBS_URL_PREFIX}/resource/{repo}/{type}/{tag}"
```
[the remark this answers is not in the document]
sameo: Ah right, I forgot that one. Fixed now, thanks.

src/kbc_modules/cc_kbc/mod.rs
sameo: We do. I'll add one.

src/kbc_modules/cc_kbc/mod.rs
sameo: I see. Would the following check:

```rust
if resource.kbs_addr.len() > 0 && kbs_host != resource.kbs_addr
```

make more sense? let me give it a go.

jialez0: > I see. Would the following check:
> 
> ```rust
> if resource.kbs_addr.len() > 0 && kbs_host != resource.kbs_addr
> ```
> 
> make more sense? let me give it a go.

I think this would be OK. Thanks!
Let me explain in detail the use of URIs in our current implementation:
AA now has two APIs, one for get resource and the other for decrypt annotation. The call inputs for these two ports are different:

For the `get resource` API, the caller is `image-rs`, which obtains the `kbc-name` and `kbs-uri` (format is like `http://1.2.3.4:8080/) from the `aa_kbc_params` (format is like `cc_kbc::http://1.2.3.4:8080/) passed in from `kata-agent` config, and obtains the Resource URI (format is like `[kbs:///a/b/c`](kbs:///a/b/c%60)) from the image validation policy file. Then `image-rs` get `resource_path` (format is like `/a/b/c`) from Resource URI, and pass `resource_path` to AA together with the `kbc-name` and `kbs uri` to call AA `get resource` interface.

For the `decrypt annotation` API, the caller is `ocicrypt-rs`. Similarly, `ocicrypt-rs` obtains the `kbc-name` and `kbs-uri `from `aa-kbc-params`, but `ocicrypt-rs` does not resolve the Resource URI contained in the annotation (format is `[kbs:///a/b/c`](kbs:///a/b/c%60)) . Instead, the annotation is directly passed to AA along with the `kbc-name` and `kbs-uri` to call AA `decrypt annotation` interface.

This difference results in different processing logic for Resource URI between the two interface of AA, specifically:

The get resource port obtains the `kbs-uri` and `resource-path` in the input parameters, initializes the KBC with the `kbs uri`, recombines the `kbs uri` and `resource path` into a REAL Resource URI (Now format is like `[kbs://1.2.3.4:8080/a/b/c`](kbs://1.2.3.4:8080/a/b/c%60)), and converts the RESL Resource URI into a real URL (format is `[http://1.2.3.4:8080/kbs/v0/a/b/c`](http://1.2.3.4:8080/kbs/v0/a/b/c%60)).

The decrypt annotation port obtains the `kbs uri `in the input parameters, and initialize the KBC using the `kbs uri`. It then parses the Resource URI in the annotation (format is `[kbs:///a/b/c`](kbs:///a/b/c%60)）And try to convert this Resource URI into a real URL (format is `[http://1.2.3.4:8080/kbs/v0/a/b/c`](http://1.2.3.4:8080/kbs/v0/a/b/c%60))

Therefore, if we throw an error here, there is no problem with `get resource` interface, but there will be an error with `decrypt annotation`.

jialez0: @sameo Sorry for the text format confusion, I have fixed these text confusion, you can refresh the page:-)

jialez0: There is another point to note, although I said above that `image-rs` and `ocictrypt-rs` get `kbs-uri` in format of `http://1.2.3.4:8080` from `aa_kbc_params`. However, due to some historical reasons, in the current code, `eaa_kbc` and `sev_kbc` require `kbs-uri` in `aa_kbc_params` must not contain a scheme prefix (`http://`)

mkulke: Thank you for that explanation. Is there a reason why `ocicrypt-rs` does not disassemble the resource URI?

jialez0: cc @Xynnn007

jialez0: It seems that I have some problems with text conversion. I fixed them. You can refresh the page again :-)

jialez0: @mkulke And the concept of ocicrypt "Key Provider" is defined here: https://github.com/containers/ocicrypt/blob/main/docs/keyprovider.md

src/kbc_modules/cc_kbc/mod.rs
sameo: Done now.

src/kbc_modules/cc_kbc/mod.rs
sameo: @jialez0 I added a few fixes. Let me know if that looks better and work for you.

src/kbc_modules/cc_kbc/mod.rs
```diff
@@ -146,7 +151,7 @@ impl Kbc {
 
         let challenge = self
             .http_client()
-            .post(format!("{kbs_uri}/{KBS_URL_PREFIX}/auth"))
+            .post(format!("{kbs_uri}{KBS_URL_PREFIX}/auth"))
```
jialez0, line 154: Why we remove `/` here?

jialez0: Does this mean the `kbs_uri` in `fn new(kbs_uri: String) -> Result<Kbc>` must be `http://.../` but can not be `http://...` ?

jialez0: Oh, I saw the test cases in the unit tests, and they worked properly. There's no problem doing this. Sorry!

jialez0 (approved):
LGTM. Huge thanks for this! @sameo

Xynnn007:
Hi @sameo @mkulke Thanks a lot for this PR and bring the problem out! Thanks @jialez0 for the very detailed explanation!

It is clear that now the KBS Resource URI scheme is still not well integrated into the overall CoCo architecture, and we might need to do some other things to make the whole design reasonable. Let me try to explain this:

#### Contradiction
1. AA **only supports one KBS** now, and the address of the KBS is given by `aa_kbc_params` from kernel commandline.
2. `<kbs_host>:<kbs_port>` field in a KBS Resource URI implies that the resource could be from different KBS.


#### Temporary Workaround

- Overwriting all the `<kbs_host>:<kbs_port>` field of a KBS Resource URI with that given by `aa_kbc_params`. In this way all confidential resource requests (`download_confidential_resource` and KEK id in `decrypt_image_layer_annotation` ) could be handled by the KBS specified by `aa_kbc_params`

In this way, (yes we are here), we can make the whole CoCo system work.

#### Problems

It is clear that the temporary workaround is just trying to fit in the current design, s.t. the KBS address is given already and this is the only KBS could be accessed. Thus the API of `download_confidential_resource` and `decrypt_image_layer_annotation` is somehow weird.

The key point is -- it is "where the resources are stored" specified by `aa_kbc_params`. IMO, it should be "who can authorize this workload to access KBS", and an AA can connect to different KBSes.

This is because a KBS for guest is not only "where the resources are store", but also "who can authorize me to access the resources" (by remote attestation).

#### Probably A Solution in future

1. decouple KBS into "confidential resource storage" part and "auth" part. Then KBS is only for confidential resource storage. Let's name the "auth" part "PS" (Provision Service).
2. Bringing in "passport model". `aa_kbc_params` should be replaced by the address of PS.
3. When AA starts, it will connect to PS and do authentication (including a underlying remote attestation). PS then authorizes a passport to the AA.
4. Then, AA can handle resource requests of different KBS Resource URIs, make requests to different KBS. All the requests are privileged by the passport authorized in step 3.

This can fit in well with the current implementation of KBS (user public key, JWT, etc.)

I think it is a systematic problem and we might need to do changes to several componants and interaction steps (based on the KBS Attestation Protocol).

We've posted a proposal for this https://github.com/confidential-containers/confidential-containers/issues/131, please take a look whenever it is convenient and give any comments and suggestions.

This could cover the problem of "identity of KBS" and "evidence factory attack". From my own point if view, we should do something to resolve this from probably 0.6.0 or 0.7.0.

cc @fitzthum

PR #173, opened 2023-03-28
Add support for Azure SNP vTPM attestation

mkulke:
Draft, feedback welcome!

This aims to add attestation code for Azure SEV-SNP CVMs. The process can be roughly depicted like this: In those CVMs the SNP Attestation Report is fetched initially within Azure's HCL subsystem with VMPL=0 privileges. To link the report to the machine, a hash of the public part of the vTPM's _Attestation Key_ (AK) is passed as `report_data` during the retrieval of the SNP report. To prove freshness, we request a vTPM AK-signed nonce'd quote, using the nonce we get as challenge from kbs and embed the resulting quote in the _TEE Evidence_. 
 
A respective `AzSnpVtpm` verifier - besides validating the SNP report - will also verify that the quote we generated has been signed by the AK and the nonce is matching.  

## Notes

At the moment this is relying on a fork of `kbs-types`, since there is no new release including the TEE entry yet.

More discussion: https://github.com/confidential-containers/kbs/issues/51)


Cargo.toml
```diff
@@ -30,6 +30,8 @@ tonic = { version = "0.8.0", optional = true }
 url = "2.3.1"
 uuid = { version = "1.1.2", features = ["serde", "v4"], optional = true }
 zeroize = "1.5.7"
+az-snp-vtpm = { version = "0.1", optional = true }
+hex = "0.4.3"
```
dubek, line 34: where is `hex` used in this PR?

src/kbc_modules/cc_kbc/attester/az_snp_vtpm/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+    report: Vec<u8>,
+}
+
+impl Attester for AzSnpVtpmAttester {
+    fn get_evidence(&self, report_data: String) -> Result<String> {
+        let report = vtpm::get_report()?;
+        let report_data_bin = base64::decode(&report_data)?;
+        let quote = vtpm::get_quote(&report_data_bin)?;
+
+        let evidence = Evidence { quote, report };
```
dubek, line 30: Are the VCEK, ASK, ARK sent here? 
Or do we expect the verifier to extract the chip ID and fetch the VCEK and chain from AMD's keys service?

dubek (commented):
Thanks for submitting this!
A few questions below:

src/kbc_modules/cc_kbc/attester/az_snp_vtpm/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use super::Attester;
+use anyhow::*;
+use az_snp_vtpm::vtpm;
+use serde::{Deserialize, Serialize};
+
+pub fn detect_platform() -> bool {
+    vtpm::has_tpm_device()
```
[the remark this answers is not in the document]
mkulke: cc @jialez0 @Xynnn007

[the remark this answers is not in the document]
Xynnn007: I'm with the idea that decide which attester is during the compilation. We can pre-build a bunch of images, s.t.
`image with tdx attester`, `image with snp attester` and `image with vtpm attester`. when deploying, the user could choose which one to deploy. 

cc @fitzthum

src/kbc_modules/cc_kbc/attester/az_snp_vtpm/mod.rs
jepio: We can fetch VCEK/ASK/ARK from the IMDS server on the underlying node and i think that would work better than the KBS having to reach out to AMD's servers.
See also: https://github.com/mkulke/attestation-service/pull/1#discussion_r1139900623.

mkulke: At the moment the verifier does fetch the VCEK + cert chain from AMD, which is not practical (there are rather rigid rate limits on that server).

There are multiple options:
- A metadata endpoints returns VCEK, ASK, ARK in Azure, the attester could include them in the report (extended snp report)
- The ARK + ASK could be part of the compiled binary (I think the nascent https://github.com/virtee/snp crate bundles the ARKs and ASK for genoa and milan)
- There is a caching facility in the sev crate (`sev::cached_chain`)

I'm leaning towards bundling VCEK, ASK in the report from the attester and caching the ARK on the verifier certs.

fitzthum: For reference the draft of the SNP attester/verifier uses the extended report to hold the VCEK (cached on the host, basically) and the ARK and ASK are in the verifier source.

mkulke: I think we'll do something similar. We'll fetch the VCEK from the metadata service (and cache it) and use the key chain bundled in the sev crate.

Xynnn007 (approved):
LGTM, thanks for the great work!

attester/Cargo.toml
```diff
@@ -13,9 +13,11 @@ serde.workspace = true
 serde_json.workspace = true
 strum.workspace = true
 tdx-attest-rs = { git = "https://github.com/intel/SGXDataCenterAttestationPrimitives", rev = "cc582e8be0c9010295c66fb58c59f74744017600", optional = true }
+az-snp-vtpm = { version = "0.2", optional = true }
```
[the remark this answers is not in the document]
mkulke: 👍 I also added a feature flag to the dep, so we avoid a transitive openssl dependency for AA

jialez0 (approved):

sameo (approved):

PR #182, opened 2023-04-11
Optimized the organizational structure of the code

jialez0:
**Note**: This PR only adjusts the code layout and does not modify or add any functionality, nor does it change any call relationships.

This PR has rearranged the code structure of the repository, organizing the code in the form of a workspace, and extracting some core modules into independent libraries to enable AA to integrate more flexibly into a wider range of scenarios.

Specifically, the following modifications have been made:
    1. Extract crypto related operations and Resource URI parsing functions into separate libraries and place them in the `deps` directory
    2. Move the original `attestation_agent` library to `lib` directory.
    3. Extract the KBC module into a separate library
    4. Extract the KBS client functional components of cc-kbc into a separate library
    5. Extract Attesters of cc-kbc into a separate library


deps/sev/Cargo.toml
fitzthum, line 1: At some point we might just want to deprecate the `offline_sev_kbc` and move this all to the online kbc. We can worry about that later.

kbs_client/Cargo.toml
fitzthum, line 1: `KbsClient` name is slightly confusing given that we also have KBCs. Can we make it more clear that this is just for the `cc-kbc` and generic KBS? I think the original name `kbs_protocol` might be more clear.

wrapper/Cargo.toml
fitzthum, line 1: The wrapper name is a little bit vague. Not sure what would be better, though.

fitzthum (commented):
A few minor comments but I think we're on the right track.

kbs_client/Cargo.toml
jialez0: OK, let's rename the `kbs_client` crate to `kbs_protocol` crate, and rename the `KbsClient` structure to `KbsProtocolWrapper`. That's would be clearer.

wrapper/Cargo.toml
jialez0: This is just the name of the directory, and the name of this crate member is still `attestation_agent`. `wrapper` may indeed be a bit vague. What if we call this directory `lib` directly?

fitzthum (approved):
LGTM. This will require rebasing all the outstanding verifier PRs, but that's ok.

PR #184, opened 2023-04-18
New tee type: CCA (Confidential Compute Architecture)

chendave:


Cargo.toml
```diff
@@ -39,7 +39,7 @@ rstest = "0.16.0"
 tonic-build = { version = "0.8.0", optional = true }
 
 [features]
-default = ["sample_kbc", "rust-crypto"]
+default = ["sample_kbc", "rust-crypto", "cc_kbc"]
```
fitzthum, line 42: Is this just for testing?

Note how other platforms have an additional feature to build their attester. You may want to add something similar.

src/kbc_modules/cc_kbc/attester/cca/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use super::Attester;
+use anyhow::*;
+use std::env;
+
+
+// If the environment variable "CCA_ATTESTER" is set,
+// the TEE platform is considered as "CCA".
+pub fn detect_platform() -> bool {
+    env::var("CCA_ATTESTER").is_ok()
```
fitzthum, line 14: Who sets this? Is there a way to query the platform more directly like via cpuid?

src/kbc_modules/cc_kbc/attester/cca/mod.rs
```diff
@@ -0,0 +1,30 @@
...
+
+
+#[derive(Debug, Default)]
+pub struct CCAAttester {}
+
+// NOTE: If we sign the evidence here rather by a veraison proxy (proxy to veraison verifier), we need to rustify the cbor lib to support the logic around signature.
+#[allow(unused_variables)]
+impl Attester for CCAAttester {
+    fn get_evidence(&self, data: String) -> Result<String> {
+        let s = std::include_str!("cca-claims-without-realm-challenge.json").as_bytes();
```
fitzthum, line 25: This is some kind of placeholder?

src/kbc_modules/cc_kbc/attester/mod.rs
```diff
@@ -15,20 +16,23 @@ pub mod tdx;
...
 /// - Sample: A dummy TEE that used to test/demo the KBC functionalities.
+/// - CCA: Arm Confidential Compute Architecture TEE.
 #[derive(Debug, EnumString, Display)]
 #[strum(ascii_case_insensitive, serialize_all = "lowercase")]
 pub enum Tee {
     Tdx,
     Sgx,
     Sevsnp,
     Sample,
+    CCA,
```
fitzthum, line 27: `CCA` -> `Cca`

fitzthum (commented):
I see this is WIP and it looks like it's just using some dummy evidence, but I got so excited to see CCA support that I had to give a few suggestions.

Also note that https://github.com/confidential-containers/attestation-agent/pull/182 will require a rather large rebase one merged.

src/kbc_modules/cc_kbc/attester/cca/mod.rs
chendave: I was thinking this might be an env, just like `AA_SAMPLE_ATTESTER_TEST`, query from platform is a great idea. Let's see if we can move toward that.

Cargo.toml
chendave: yes, this will be reverted I think.

src/kbc_modules/cc_kbc/attester/cca/mod.rs
chendave: currently, I don't have a elegant way to get back the evidence, so just load it in this way instead, this will be updated if we there is any progress for us to get back the real evidence.

chendave:
> I see this is WIP and it looks like it's just using some dummy evidence, but I got so excited to see CCA support that I had to give a few suggestions.

@fitzthum thanks a lot for your review, any comments is highly appreciated!

PR #189, opened 2023-04-23
KMS: add API definitions and crate for KMS integration

Xynnn007:
This PR include the following:
- Added a crate `deps/kms`
- Added Datakey API abstraction inside `deps/kms`
- Added KMS module API abstraction inside `deps/kms`
- Implemented an example for KMS

related to https://github.com/confidential-containers/guest-components/issues/188

This PR will help to give a basic lib for image en/decryption using KMS.

cc @fitzthum @jialez0

### DIFF
--- a/deps/kms/src/api.rs
+++ b/deps/kms/src/api.rs
@@ -1,0 +1,29 @@
+// Copyright (c) 2023 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+use anyhow::Result;
+use async_trait::async_trait;
+
+#[async_trait]
+pub trait KMS {
+    /// Generate a key inside the KMS, and return an unique id string
+    /// that represents the key inside KMS. Then this unique id string
+    /// can be used to encrypt data slice or decrypt ciphertext that
+    /// has been encrypted with the key.
+    ///
+    /// Note: the returned string can only allow letters, numbers,
+    /// underscores `_` and hyphen `-` are allowed
+    async fn generate_key(&mut self) -> Result<String>;
+
+    /// Use the key of `keyid` to encrypt the `data` slice inside KMS, and then
+    /// return the ciphertext of the `data`. The encryption operation should occur
+    /// inside KMS. This function only works as a wrapper for different KMS APIs
+    async fn encrypt(&mut self, data: &[u8], keyid: &str) -> Result<Vec<u8>>;
+
+    /// Use the key of `keyid` to decrypt the `ciphertext` slice inside KMS, and then
+    /// return the plaintext of the `data`. The decryption operation should occur
+    /// inside KMS. This function only works as a wrapper for different KMS APIs
+    async fn decrypt(&mut self, ciphertext: &[u8], keyid: &str) -> Result<Vec<u8>>;
+}
--- a/deps/kms/src/lib.rs
+++ b/deps/kms/src/lib.rs
@@ -1,0 +1,22 @@
+// Copyright (c) 2023 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! # KMS
+//!
+//! This lib defines the API of the KMSes which will be used to implement
+//! encryption and decryption.
+//!
+//! ## Trait Definitions
+//!
+//! ### KMS
+//!
+//! [`KMS`] defines the interface of KMS.
+
+pub mod api;
+pub use api::KMS;
+
+pub mod plugins;
+
+pub mod types;
--- a/deps/kms/src/plugins/sample.rs
+++ b/deps/kms/src/plugins/sample.rs
@@ -1,0 +1,111 @@
+// Copyright (c) 2023 Alibaba Cloud
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+//! This is a sample KMS implementation example.
+//!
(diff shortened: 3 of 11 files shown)


deps/kms/src/plugins/sample.rs
```diff
@@ -0,0 +1,127 @@
...
+//
+
+use aes_gcm::{aead::Aead, Aes256Gcm, Key, KeyInit, Nonce};
+use anyhow::{anyhow, Result};
+use async_trait::async_trait;
+use crypto::WrapType;
+use rand::RngCore;
+use zeroize::Zeroizing;
+
+use crate::{Datakey as DatakeyTrait, KMS};
```
dubek, line 13: The alias is a bit confusing, because you later define a `pub struct Datakey`. I suggest:

1. `use crate::{Datakey, KMS}`
2. line in 34: `pub struct SampleDatakey` (and then later: `impl Datakey for SampleDatakey`)

Xynnn007: Thanks for the suggestion. Fixed it

jepio (commented):
I don't understand the "datakey" abstraction in the context of AA/KBS. Can you explain how it is going to be used?

deps/kms/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+//! ## Trait Definitions
+//!
+//! ### Datakey
+//!
+//! [`Datakey`] is an abstraction of a key with the following attributes
+//! in different KMSes:
+//! - Can be used to encrypt/decrypt large pieces of data
+//! - Plaintext of the key can be exported outside the KMS
+//! - KMS can help to decrypt the ciphertext of the key to get the plaintext
+//! of the key.
```
jepio, line 20: why do we need the KMS to "decrypt the ciphertext of the key", if "the plaintext of the key can be exported outside the KMS".

i thought keys were going to be wrapped using a TEE pubkey.

deps/kms/src/plugins/sample.rs
```diff
@@ -0,0 +1,127 @@
...
+        Self { key }
+    }
+}
+
+#[async_trait]
+impl Datakey for SampleDatakey {
+    async fn encrypt(&mut self, plaintext: &[u8]) -> Result<Vec<u8>> {
+        let encryption_key = Key::<Aes256Gcm>::from_slice(&self.key);
+        let cipher = Aes256Gcm::new(encryption_key);
+        let nonce = Nonce::from_slice(IV);
```
jepio, line 55: This is kind of a bad example, because the IV needs to be returned together with the ciphertext.
And decrypt needs to receive the IV.

deps/kms/src/plugins/sample.rs
```diff
@@ -0,0 +1,127 @@
...
+use crypto::WrapType;
+use rand::RngCore;
+use zeroize::Zeroizing;
+
+use crate::{Datakey, KMS};
+
+const IV: &[u8] = b"unique nonce";
+
+/// Only for sample
+pub const EXAMPLE_CMK: &[u8] = &[
```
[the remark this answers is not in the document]
Xynnn007: It is Customer Master Key abstraction. Different KMS will call it a different name. The core idea of this key is that it never leave KMS in plaintext.

deps/kms/src/api.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Use this key to encrypt the given data slice, and return the
+    /// ciphertext blob that can be decrypted by calling `decrypt` API.
+    async fn encrypt(&mut self, plaintext: &[u8]) -> Result<Vec<u8>>;
+
+    /// Use this datakey to decrypt the given data blob, and return the
+    /// plaintext.
+    async fn decrypt(&mut self, ciphertext: &[u8]) -> Result<Vec<u8>>;
+
+    /// Export the key blob that can be used to reconstruct the Datakey
+    /// except the plaintext of the Datakey.
```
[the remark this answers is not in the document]
Xynnn007: In my mind the key blob will be stored in KBS. When the datakey is to be used, KBS can use this blob to reconstruct the plaintext of this datakey. But the plaintext cannot appear when the key blob is being stored. Usually the ciphertext of the datakey, the CMK (a.k.a key encryption key) id to encrypt this datakey, etc are in the key blob.

deps/kms/src/lib.rs
Xynnn007: This design philosophy comes from the API design from [aliyun KMS](https://www.alibabacloud.com/help/en/key-management-service/latest/kms-generatedatakey), [aws KMS](https://docs.aws.amazon.com/kms/latest/APIReference/API_GenerateDataKey.html) etc. It is similiar to "Envelope encryption", s.t. wrap the datakey with CMK.

Xynnn007:
Hi @jepio , I've updated the PR to try to simplify the API. Also, `iv` is attached in Sample KMS. Thanks for your suggestions, please take another look if it looks good to you.

jepio:
So the interface now is:
 - `generate_key` generates a symmetric KEK
 - `encrypt` is used for LEK encryption (wrapping)
 - `decrypt` is used for LEK decryption (unwrapping)
?

If this is only for image decryption/encryption then that's probably good enough, but we'll need to see how the API fits existing implementations.

Xynnn007:
@jepio This is a basic abstraction for KMSes. We can use this as a primitive to construct the image en/decryption. After this PR I will open another to impl this.

PR #199, opened 2023-05-12
KBS protocol: Add function to get attestation token

jialez0:


kbs_protocol/src/lib.rs
```diff
@@ -177,6 +181,18 @@ impl KbsRequest for KbsProtocolWrapper {
 
         bail!("Request KBS: Attested but KBS still return Unauthorized")
     }
+
+    async fn attest(&mut self, url_string: String) -> Result<String> {
+        let url = Url::parse(&url_string)
```
jepio, line 186: why do we need to parse the url and not pass it directly to `self.attestation`? I don't know what you imagine the caller of this function would do.

jepio (commented):
This is simple enough, but I don't know whether this is useful just by looking at the interface. Where do you want to call this from?

jialez0:
@jepio We are gradually increasing the ability to use passport models for the attestation system, this API is designed to serve passport model scenarios. When the caller needs to obtain an attestation token, they can directly call this API without worrying about the complex evidence acquisition process and HTTPS interaction process.

jialez0:
> This is simple enough, but I don't know whether this is useful just by looking at the interface. Where do you want to call this from?

Now, the client tool of KBS is using this interface: https://github.com/confidential-containers/kbs/blob/main/tools/client/src/main.rs#L99, and in the near future, AA will also provide a public API for upper level applications to obtain attestation token.

kbs_protocol/src/lib.rs
```diff
@@ -21,7 +21,10 @@ pub const KBS_URL_PREFIX: &str = "kbs/v0";
 
 #[async_trait]
 pub trait KbsRequest {
-    async fn http_get(&mut self, url: String) -> Result<Vec<u8>>;
+    /// Get confidential resource
+    async fn http_get(&mut self, resource_url: String) -> Result<Vec<u8>>;
+    /// Attestation and get attestation results token (Base64 endcoded)
```
wainersm, line 26: s/endcoded/encoded/

kbs_protocol/src/lib.rs
```diff
@@ -113,7 +116,8 @@ impl KbsProtocolWrapper {
         match attest_response.status() {
             reqwest::StatusCode::OK => {
                 self.authenticated = true;
-                Ok(())
+                let token = attest_response.json::<serde_json::Value>().await?["token"].to_string();
```
wainersm, line 119: Would make sense to check here the token is really base64 encoded?

jepio: the token is supposed to be a json-web-token which has base64 components but is simply a base64 encoded string.

mythi:
KBS's `tools/client` depends on this so it'd be nice to have this in the release so that `tools/client` can be updated.

jialez0:
@sameo Could you PTAL?

jepio (approved):

sameo (approved):

PR #204, opened 2023-05-17
attester: Remove old occlum attester

fitzthum:
I think we can remove this file, right? @Xynnn007 @jialez0


Xynnn007 (approved):
Thanks! Sorry for this. This should be removed.

jialez0 (approved):

PR #212, opened 2023-06-03
Rename Occlum attester to SGX attester and add Gramine support to it

mythi:
WIP: I'm planning to change the existing `occlum-attester` to `sgx-dcap-attester`.

TODO:

- [x] rename feature names
- [x] rename structs/modules
- [x] rebase this PR

### DIFF
--- a/attester/src/sgx_dcap/mod.rs
+++ b/attester/src/sgx_dcap/mod.rs
@@ -10,8 +10,29 @@ use serde::{Deserialize, Serialize};
 
 const OCCLUM_ENV: &str = "OCCLUM";
 
+enum SgxLibOsType {
+    Invalid,
+    Occlum,
+    Gramine,
+}
+
+fn get_libos_type() -> SgxLibOsType {
+    if std::env::var(OCCLUM_ENV).is_ok() {
+        return SgxLibOsType::Occlum;
+    }
+
+    match std::fs::read_to_string("/dev/attestation/attestation_type") {
+        Ok(d) if d == "dcap" => SgxLibOsType::Gramine,
+        _ => SgxLibOsType::Invalid,
+    }
+}
+
 pub fn detect_platform() -> bool {
-    std::env::var(OCCLUM_ENV).is_ok()
+    match get_libos_type() {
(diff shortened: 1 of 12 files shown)


mythi:
/cc @Xynnn007 

I've tested both Gramine and Occlum paths manually. Occlum tester is also fixed and can be added back to CI.

attester/src/sgx_dcap/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        SgxLibOsType::Invalid => false,
+        SgxLibOsType::Occlum => true,
+        SgxLibOsType::Gramine => true,
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+struct SgxDcapAttesterEvidence {
+    /// Base64 encoded SGX quote.
+    quote: String,
```
Xynnn007, line 41: Do we need to add another field here named `libos_type` to indicate the type of libos? Though the integrity of this field is not protected.

Xynnn007 (approved):
LGTM and I like the reorganization.

attester/src/sgx_dcap/mod.rs
mythi: In my first version I had it for `SgxDcapAttester` and then `Defaulter` implemented for it that initialized it but then dropped that version. The verifier can use `quote` regardless of who generated it.

Xynnn007: Well. I'm thinking about how the verifier choose the golden value of the MRENCLAVE.
1. With extra `libos_type`: There should be two lists of golden values for occlum and gramine agent-enclave, and they can be queried due to different `libos_type`
2. We only keep one list of both golden values for occlum and gramine agent-enclave. When verifying the quote, attestation service will use the values in the list to try to match the one in quote.

Both are fine with me.

mythi: OK I see what you mean. I'd keep this PR as it is now and think about the quote verification for Gramine/Occlum co-existence later.

mythi:
moved to `image-rs`

PR #214, opened 2023-06-08
kbs_protocol: Return unquoted string from attestation()

jepio:
Currently the String returned from attestation() is quoted, because Value["token"] is a Value, and to_string() returns the "string representation" of the Value instead of the contained string. Introduce a struct representing the data returned from /attest, and deserialize into that so that we return the correct, raw string.


jepio:
cc @jialez0

PR #219, opened 2023-06-16
Adjusted the handling of URLs to support more types of hosts

jialez0:
Currently, our default host URL format for KBS is http://IP:PORT, But in productization, we may give host URL more forms. For example: `http://my-kbs.com` , `http://my-kbs.com/user-alice`. Therefore, it is necessary to expand the supported KBS host URL format, More precisely, it should no longer be referred to as the "host URL", but rather as the "root URL".


kbs_protocol/src/lib.rs
```diff
@@ -136,19 +136,16 @@ impl KbsProtocolWrapper {
...
     async fn http_get(&mut self, url_string: String) -> Result<Vec<u8>> {
-        let url = Url::parse(&url_string)
+        let _ = Url::parse(&url_string)
             .map_err(|e| anyhow!("Invalid Request URL {url_string}: {e}"))?;
 
-        // Use default port of KBS: 8080
-        let port = url.port().unwrap_or(8080);
-        let host_str = url.host_str().ok_or_else(|| anyhow!("No KBS host given"))?;
-        let kbs_url = format!("{}://{host_str}:{port}", url.scheme());
+        let root_url = url_string.split(KBS_URL_PREFIX).collect::<Vec<&str>>()[0].to_string();
```
Xynnn007, line 142: Is it possible that the provided `url_string` does not have a substr `KBS_URL_PREFIX`? Is the returned vector as expected?

PR #220, opened 2023-06-20
attester: az_snp_vtpm: Check cpuid bits in detect_platform()

jepio:
It took a while to realize that we can check cpuid registers exposed by Hyper-V to determine whether we are in an SNP VM that can use the az-snp-vtpm attester. Rework detect_platform() to perform these checks.


attester/src/az_snp_vtpm/mod.rs
```diff
@@ -7,11 +7,44 @@ use super::Attester;
...
+            if hyper_info.identify() != Hypervisor::HyperV {
+                debug!("vTPM is only supported on Hyper-V");
+                return false;
+            }
+        }
+        None => {
+            debug!("Not a VM");
+            return false;
+        }
+    }
```
mkulke, line 32: Since we pass through the detection heuristics for each attester module (i.e. it "correctly" fails on other CSPs) this could add misleading output. Maybe we can make it less error-ish. (vTPMs also exist on GCP, afaik)

```suggestion
    let Some(hyper_info) = cpuid.get_hypervisor_info() else {
        debug!("Not a VM");
        return false;
    }   
    let hv_platform = hyper_info.identify();
    debug!("Hypervisor: {:?}", hv_platform;
    if hv_platform != Hypervisor::HyperV {
        return false;   
    }    
```

mkulke (approved):

